PR #31, opened 2018-11-20
Update SetPedStrafeClipset.md

Tianshee:
Defines movement when _aiming_.


Tianshee:
Like this?

Tianshee:
Or ``` must be removed everywhere?

Tianshee:
Or maybe like this...

smallo92:
You had it right the second commit. The 3rd one is wrong, and you could have easily seen that by looking at any other previously accepted PRs.

Tianshee:
Some accepted commits have even multiple ``` blocks.

PR #37, opened 2018-11-25
Update arguments on TaskJump

Xinerki:
Add some more info too.


blattersturm:
Adding arguments/changing argument types is not currently viable due to C# API class wrapping.

PR #46, opened 2018-11-30
Update SetEntityAsNoLongerNeeded.md

Tianshee:
Added important note.

### DIFF
--- a/ENTITY/SetEntityAsNoLongerNeeded.md
+++ b/ENTITY/SetEntityAsNoLongerNeeded.md
@@ -8,10 +8,9 @@ ns: ENTITY
 void SET_ENTITY_AS_NO_LONGER_NEEDED(Entity* entity);
 ```
 
-```
-Marks the specified entity (ped, vehicle or object) as no longer needed.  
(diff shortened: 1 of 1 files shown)


Tianshee:
You checked and didn't get same result or just biased and didn't read anything?

Tianshee:
It's either a bug or lack of documentation. And FiveM geniuses that are always right are absolutely sure that it's not a bug. So documentation should be updated. Right? No, you declined that addition. Then it's not a bug, not lacking documentation. What then?

Bluethefurry:
You removed the ```quotations```, also, the information is incorrect and/or useless, SetEntityAsNoLongerNeeded works without any special checks or prerequisites.

blattersturm:
This is not the place for discussing drama. Locked.

If you wish to discuss this, do so on the forums, on Discord, or using email to support@fivem.net.

In addition, the documentation is _not_ the place for placing workarounds to perceived issues, perceived issues should be fixed using a pull request to the FiveM repository or reported as a well-written and objective report with a repro.

PR #56, opened 2019-01-01
Some informations about 0x83f813570ff519de native

FumaPraQue:
I'm already using this native for a long time, and it's working properly.


DisableVehicleNeonLights.md
```diff
@@ -0,0 +1,15 @@
+---
+ns: VEHICLE
+aliases: ["N_0x83f813570ff519de"]
```
blattersturm: This alias type shouldn't have a N_ prefix.

FumaPraQue: Fixed.

PR #58, opened 2019-01-08
Add new a native: SetVehicleHeadlightsColor

PichotM:
TITLE

### DIFF
--- a/AUDIO/OverrideVehHorn.md
+++ b/AUDIO/OverrideVehHorn.md
@@ -5,8 +5,30 @@ ns: AUDIO
 
 ```c
 // 0x3CDC1E622CCE0356 0x2ACAB783
-void OVERRIDE_VEH_HORN(Vehicle vehicle, BOOL mute, int p2);
+void OVERRIDE_VEH_HORN(Vehicle vehicle, BOOL override, int hornHash);
 ```
+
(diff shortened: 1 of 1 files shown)


VEHICLE/SetVehicleHeadlightsColor.md
```diff
@@ -4,7 +4,7 @@ ns: VEHICLE
 ## SET_VEHICLE_HEADLIGHTS_COLOR
 
 ```c
-// 0xE41033B25D003A07
+// _0xE41033B25D003A07
```
TomGrobbe: I don't think that's what he meant, he meant the custom name you've given it should be prefixed with `_`. The hash there shouldn't.

Not sure, but I think this is also required (put it below the ns: VEHICLE line, but above the --- ) 
```
aliases: ["0xE41033B25D003A07"]
```

PichotM: ty

TomGrobbe:
Just noticed something, might be worth changing it to `...Colour` instead of `Color` since Rockstar tends to use `Colour` instead of `Color` in their native names. 🤷‍♂️

PichotM:
yep, why not do you wanna also add the Get native?

PR #69, opened 2019-02-17
Document SetMinimapComponent

PichotM:
- Description of the native
- An incomplete list of minimap's components.

### DIFF
--- a/HUD/SetMinimapComponent.md
+++ b/HUD/SetMinimapComponent.md
@@ -5,7 +5,7 @@ ns: HUD
 
 ```c
 // 0x75A9A10948D1DEA6 0x419DCDC4
-Any SET_MINIMAP_COMPONENT(int p0, BOOL p1, int p2);
+void SET_MINIMAP_COMPONENT(int componentID, BOOL enable, int p2);
 ```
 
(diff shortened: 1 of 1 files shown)


HUD/SetMinimapComponent.md
```diff
@@ -5,17 +5,28 @@ ns: HUD
 
 ```c
 // 0x75A9A10948D1DEA6 0x419DCDC4
-Any SET_MINIMAP_COMPONENT(int p0, BOOL p1, int p2);
+void SET_MINIMAP_COMPONENT(int componentID, BOOL toggle, int p2);
```
blattersturm: Changing return values will have arbitrary ABI consequences. Are you sure no compiled C# code is using this?

If not (which, fyi, it's impossible not to be), _please retain the original type signature_.

HUD/SetMinimapComponent.md
```diff
@@ -5,17 +5,28 @@ ns: HUD
...
 ```
 
 ```
```
blattersturm: Why keep the old code block description (which says _nothing_ useful) on top?

PR #71, opened 2019-02-17
document SRL natives

PichotM:
- description of the natives
- srl list
- example

### DIFF
--- a/STREAMING/PrefetchSrl.md
+++ b/STREAMING/PrefetchSrl.md
@@ -8,7 +8,22 @@ ns: STREAMING
 void PREFETCH_SRL(char* srl);
 ```
 
+This native is used to attribute the SRL that BeginSrl is going to load. This is usually used for 'in-game' kinematics (not cutscenes but camera stuff) instead of SetFocusArea because it loads a specific area of the map
+which is pretty useful when the camera moves from distant areas.
+For instance, GTA:O opening cutscene.
+
+https://pastebin.com/2EeKVeLA : a list of SRL found in srllist.meta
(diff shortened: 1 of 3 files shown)


STREAMING/PrefetchSrl.md
```diff
@@ -8,7 +8,22 @@ ns: STREAMING
 void PREFETCH_SRL(char* srl);
 ```
 
+This native is used to attribute the SRL that BeginSrl is going to load. This is usually used for 'in-game' kinematics (not cutscenes but camera stuff) instead of SetFocusArea because it loads a specific area of the map
```
blattersturm: kinematics -> cinematics?

STREAMING/PrefetchSrl.md
```diff
@@ -8,7 +8,22 @@ ns: STREAMING
...
+which is pretty useful when the camera moves from distant areas.
```
blattersturm: Why is this randomly on a new line?

STREAMING/PrefetchSrl.md
```diff
@@ -8,7 +8,22 @@ ns: STREAMING
...
+For instance, GTA:O opening cutscene.
+
+https://pastebin.com/2EeKVeLA : a list of SRL found in srllist.meta
+https://pastebin.com/zd9XYUWY : here is the content of a SRL file opened with codewalker.
 
 ## Parameters
-* **srl**: 
+* **srl**: A SRL string.
```
blattersturm: `string` seems very specific - how about `name`?

PR #78, opened 2019-03-01
Update HighlightPlacementCoords.md

TomGrobbe:


TomGrobbe:
I have tried to find the real native name, but so far no luck. No idea what the first part of this native should even be called.

blattersturm:
These are `scripted glows` from `common:/data/scriptedglow.xml`, which only has 5 of them defined. Any higher values are probably random memory garbage.

TomGrobbe:
Alright well nvm then. Useless to just fix 1 mistake in the description then.

PR #86, opened 2019-03-10
document: 0x2BC54A8188768488

PichotM:


TomGrobbe:
Also heading parameter description is missing the name, you put float instead of heading.

PichotM:
Fixed here too

TomGrobbe:
👍🏻 Looks good to me

PR #97, opened 2019-03-16
Updated information on SetBlipCategory

glitchdetector:
Plus example, legend label information etc.


stannum-cfx:
This won't work properly, description text should be above the `## Parameters`, `## Examples`, `## Return value`, etc. sections. Not all over the place. It just won't be detected by the native docs tool.

PR #106, opened 2019-04-05
Minor update information on GetVehiclePedIsIn

MartenM:
Probably know by a lot of people but useful to include nonetheless (since I had to test it for my self).

### DIFF
--- a/PED/GetVehiclePedIsIn.md
+++ b/PED/GetVehiclePedIsIn.md
@@ -10,7 +10,9 @@ Vehicle GET_VEHICLE_PED_IS_IN(Ped ped, BOOL lastVehicle);
 
 ```
 Gets the vehicle the specified Ped is/was in depending on bool value.  
-[False = CurrentVehicle, True = LastVehicle]  
+[False = CurrentVehicle, True = LastVehicle]
+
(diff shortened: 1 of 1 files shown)


PED/GetVehiclePedIsIn.md
```diff
@@ -10,7 +10,9 @@ Vehicle GET_VEHICLE_PED_IS_IN(Ped ped, BOOL lastVehicle);
 
 ```
 Gets the vehicle the specified Ped is/was in depending on bool value.  
-[False = CurrentVehicle, True = LastVehicle]  
+[False = CurrentVehicle, True = LastVehicle]
+
+Returns 0 if the ped is/was not in a vehicle.
```
stannum-cfx: Add return value info to the `## Return Value` section instead.

stannum-cfx: If you're going to update a description like this, remove the code block and also document the parameters properly using the parameters section. Leave the description for explaining what the native does.

MartenM:
Ah, sorry for not using the right formatting. Finding a good example takes some time ;)

Hopefully this is better, feel free to request changes.

PR #111, opened 2019-04-21
Added documentation on native for drawing Penned In spheres

IllusiveTea:
This should hopefully A) Not break the reference and B) Add some documentation for a native which draws the spheres seen in Penned In. Maybe @TomGrobbe can wizard up a name for it :p

### DIFF
--- a/GRAPHICS/N_0x799017f9e3b10112.md
+++ b/GRAPHICS/N_0x799017f9e3b10112.md
@@ -5,17 +5,17 @@ ns: GRAPHICS
 
 ```c
 // 0x799017F9E3B10112
-void _0x799017F9E3B10112(Any p0, Any p1, Any p2, Any p3, Any p4, Any p5, Any p6, Any p7);
+void _0x799017F9E3B10112(float x, float y, float z, float radius, int r, int g, int b, float opacitiy);
 ```
-
+Draws the sphere typically seen in the GTA:O freemode event, Penned In. Example: [Image](https://images.illusivetea.me/6654p9.png)
 
 ## Parameters
-* **p0**: 
-* **p1**: 
-* **p2**: 
-* **p3**: 
-* **p4**: 
-* **p5**: 
-* **p6**: 
-* **p7**: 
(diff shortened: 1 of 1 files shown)


GRAPHICS/N_0x799017f9e3b10112.md
```diff
@@ -5,17 +5,17 @@ ns: GRAPHICS
 
 ```c
 // 0x799017F9E3B10112
-void _0x799017F9E3B10112(Any p0, Any p1, Any p2, Any p3, Any p4, Any p5, Any p6, Any p7);
+void _0x799017F9E3B10112(float x, float y, float z, float radius, int r, int g, int b, float opacitiy);
```
blattersturm: `opacitiy` -> `opacity`?

TomGrobbe:
>Maybe @TomGrobbe can wizard up a name for it :p  

It's definitely `DRAW_m/o/p....` something. but since there's no joaat hash we can't confirm or bruteforce a name.

GRAPHICS/N_0x799017f9e3b10112.md
bladecoding: Isn't changing parameter types an ABI breaking change? I'm looking for some clarification before I submit a PR on a separate unnamed native that is using Any instead of Float.

vecchiotom:
this includes an image hosted on a personal server, wouldn't it be better to upload the image to a service such as imgur or wew.wtf? that way it's chances of getting lost and becoming a dead link would be greatly reduced...

GRAPHICS/N_0x799017f9e3b10112.md
TomGrobbe: Honestly depends on the native, if it requires floats to work, and it's an undocumented native chances are slim that anyone is actually using it in scripts that break like that. Instead if it requires floats instead of ints then usually in languages like C#, you'd be manually invoking the native anyway because there's no other way to do this.

So yes it is a backwards compatibility issue in most cases, but it depends on the native.

PR #116, opened 2019-05-05
Add SET_WARNING_MESSAGE_LIST_ROW and parameter changes for SetWarningMessageWithHeader

cerium-cfx:
* Rename of *0x0c5a80a9e096d529* to *SET_WARNING_MESSAGE_LIST_ROW*. Research for correct usage not completed, but examples present.
* *SetWarningMessageWithHeader* had some incorrect parameters. Also added an example.


stannum-cfx:
Reverted because `SET_WARNING_MESSAGE_LIST_ROW` needs to be prefixed with `_` because the joaat hash is unknown.

PR #119, opened 2019-05-05
Documentation on various undocumented HUD related natives

pongo1231:


HUD/ShowSocialClubBannedScreen.md
```diff
@@ -0,0 +1,11 @@
+---
+ns: HUD
```
Disquse: aliases?

HUD/N_0x17ad8c9706bdd88a.md
```diff
@@ -8,6 +8,22 @@ ns: HUD
 void _0x17AD8C9706BDD88A(Any p0);
 ```
 
+Related to notification color flashing, setting p0 to 0 invalidates a `_SET_NOTIFICATION_FLASH_COLOR` call for the target notification.
+
+```c
```
stannum-cfx: iirc, only 1 `c` codeblock can be used, and that's the one you see above this one. This code block should not be formatted explicitly using the `c` code style, just leave the highlighting out of this one.

PR #124, opened 2019-05-12
Document N_0x7669f9e39dc17063

bladecoding:


blattersturm:
Might be a good idea to make up a temporary `_` name and add the hash as alias?

PR #147, opened 2019-07-17
Update NetworkCreateSynchronisedScene.md

PichotM:


PichotM:
according to decompiled scripts p8 is a bool
once set to true, the synced scene is looped

PichotM:
p10 is animTime

PichotM:
according to @throwarray p7 is useOcclusionPortal

PR #149, opened 2019-07-17
Update GetPlayerIdentifier.md

FuTTiiZ:


scandula-c:
The indexes aren't static so this isn't an accurate PR. Providing the _potential_ identifiers would be better imo.

IllusiveTea:
This isnt true though, as if an identifier is missing, the positions of the identifiers change. E.G. Missing steam will mean license is 0 and not steam.

cerium-cfx:
IllusiveTea is correct. The identifier array changes depending on the available identifiers for a certain player.

PR #150, opened 2019-07-19
Update vehicle door lock status doc

aivarsomelar:
Update what this function is returning and what returned int values mean


PichotM:
@GottfriedLeibniz documented this
```
VehicleLockStatus = {
    None = 0,
    Unlocked = 1,
    Locked = 2,
    LockedForPlayer = 3,
    StickPlayerInside = 4, -- Doesn't allow players to exit the vehicle with the exit vehicle key.
    CanBeBrokenInto = 7, -- Can be broken into the car. If the glass is broken, the value will be set to 1
    CanBeBrokenIntoPersist = 8,
    CannotBeTriedToEnter = 10
}
```

PR #151, opened 2019-07-23
Update N_0xe6a9f00d4240b519.md

PichotM:


PichotM:
in fact, it does not make the scaleform bigger
it adjusts the size of the scaleform to fit it in the render target

PR #153, opened 2019-07-26
Update and rename N_0xce6294a232d03786.md to GetWorldRotationOfEntity…

PichotM:
…Bone.md
It sounds weird but this is what it does.
```lua
local boneIndex = GetEntityBoneIndexByName(rouletteHandle, "Chair_Base_01")
local dunno = N_0xce6294a232d03786(rouletteHandle, boneIndex)
-> vec3(0.0, 0.0, -45.002)
```
This returns the rotation of the chair.


PichotM:
Also a new native `0xBD8D32550E5CEBFE` does the same thing but seems only used for the roulette bone?

PR #161, opened 2019-08-03
document new server natives

PichotM:


blattersturm:
apparently they fuck the build so hard the compiler crashes, gg

PR #164, opened 2019-08-20
Maximum float

Yadiiiig:
Used multiple floats that are higher then 1.0. Amount of vehicles keeps being the same.

### DIFF
--- a/VEHICLE/SetVehicleDensityMultiplierThisFrame.md
+++ b/VEHICLE/SetVehicleDensityMultiplierThisFrame.md
@@ -8,15 +8,14 @@ ns: VEHICLE
 void SET_VEHICLE_DENSITY_MULTIPLIER_THIS_FRAME(float multiplier);
 ```
 
-```
-• Usage  
-→ Use this native inside a looped function.  
-→ Values:  
-   → 0.0 = no vehicles on streets  
-   → 1.0 = normal vehicles on streets
-----
-1.0 Seems to be the maximum.
-```
+**Usage:**
+
+- Use this native inside a looped function.
+- Values:
+  - `0.0` = no vehicles on streets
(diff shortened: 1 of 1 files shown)


gottfriedleibniz:
From the looks of it, SetSomeVehicleDensityMultiplierThisFrame simply chains together SetRandomVehicleDensityMultiplierThisFrame and SetParkedVehicleDensityMultiplierThisFrame.

```c
void __fastcall SetVehicleDensityMultiplierThisFrame(float a1)
{
  if ( *(float *)&randomVehicleDensity >= a1 )
    randomVehicleDensity = LODWORD(a1);
  if ( *(float *)&parkedVehicleDensity >= a1 )
    parkedVehicleDensity = LODWORD(a1);
}
```

Also, _SetSomeVehicleDensityMultiplierThisFrame is the only vehicle density native that allows values greater than 1.0.

blattersturm:
> SetSomeVehicleDensityMultiplierThisFrame

SET_AMBIENT_VEHICLE_RANGE_MULTIPLIER_THIS_FRAME?

FumaPraQue:
I can confirm SetSomeVehicleDensityMultiplierThisFrame really set Ambient Vehicle Range, i use that native for a long time, you can see better if you set the range in 0.1. The ambient cars will spawn too close of you.

gottfriedleibniz:
Confirmation confirmed.

stannum-cfx:
> > SetSomeVehicleDensityMultiplierThisFrame
> 
> SET_AMBIENT_VEHICLE_RANGE_MULTIPLIER_THIS_FRAME?

Implemented in: https://github.com/citizenfx/natives/commit/4378acd7014de66fe559a2ce8cfb42fcb2b8b4af

PR #187, opened 2019-09-28
cfx/onesync: add new vehicle appearance natives

PichotM:


PichotM:
(this has been pushed)

PR #193, opened 2019-10-08
Note this native is for OneSync only

Abbbbbbbbbbbhi:
Not everyone uses OneSync Alpha.


Disquse:
There's a lot natives which uses OneSync API.

blattersturm:
Useless note.

PR #196, opened 2019-10-10
Update SetPlayerHealthRechargeMultiplier.md

Mythyzyn:
Update description

### DIFF
--- a/PLAYER/SetPlayerHealthRechargeMultiplier.md
+++ b/PLAYER/SetPlayerHealthRechargeMultiplier.md
@@ -8,20 +8,15 @@ ns: PLAYER
 void SET_PLAYER_HEALTH_RECHARGE_MULTIPLIER(Player player, float regenRate);
 ```
 
-```
-Enable or disable auto regen 
-1.0 (default, regen enable))
-0.0 (regen disable)
+## Examples
+```lua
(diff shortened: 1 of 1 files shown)


PLAYER/SetPlayerHealthRechargeMultiplier.md
```diff
@@ -8,6 +8,18 @@ ns: PLAYER
...
+Enable or disable auto regen 
+1.0 (default, regen enable))
+0.0 (regen disable)
+
+Be sure to use on a player and not on a player like ped
+(local PlayerAsPed = GetPlayerPed(-1) --// <-- NOT OK)
+
+To disable regen :
+local PlayerAsPlayer = PlayerId() --// <-- OK
+SetPlayerHealthRechargeMultiplier(PlayerAsPlayer,0.0)
```
cerium-cfx: This is already implied by the native name and first parameter and is therefore unnecessary.

Mythyzyn: Ok but I imagined that a kind of doc is better than no doc, everyone don't understand very well english and more information and some example could be nice to have. Even me who's ok with english, just to well remember what natives do, I wanted to add doc here instead of keep informations on my personal wiki

PR #211, opened 2019-11-08
Update: SetBlipRouteColour

AntwanR942:
Added all colours as far as I know. It appears after 84 they just use ID 72.
Also added text for parameters.
The colours may also apply to standard blips however further testing is needed.


nimoalr:
Maybe the best option is to just put a link to https://docs.fivem.net/game-references/blips/#blip-colors since there's already all the colors available here and seems to be the same that you added ;)

manups4e:
This is really not necessary

stannum-cfx:
Like stated above, the blip GPS route colors are exactly the same as the blip colors themselves. Which can already be found [here](https://docs.fivem.net/game-references/blips/#blip-colors). Instead of adding links to all colors, I suggest you just add the link to the FiveM docs in the description of this native if you feel like there needs to be a list of colors for this native.

PR #245, opened 2020-01-20
Update TaskPlayAnimAdvanced.md

JiminyKroket:
adjust speed variables to more defined description

### DIFF
--- a/BRAIN/TaskPlayAnimAdvanced.md
+++ b/BRAIN/TaskPlayAnimAdvanced.md
@@ -9,7 +9,7 @@ void TASK_PLAY_ANIM_ADVANCED(Ped ped, char* animDict, char* animName, float posX
 ```
 
 ```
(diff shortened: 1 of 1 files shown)


Disquse:
"animEnterSpeed" and "animExitSpeed" can be bigger than 1.0 and smaller than -1.0. Even R* sometimes using it as 4.0 for enter and -4.0 as exit.

JiminyKroket:
I did not notice any difference in the speeds.  I suppose it could depend
on how fast the anim is to start. I tested with dict = 'anim@amb
@machinery@lathe@' anim = 'clean_surface_02_amy_skater_01'

JiminyKroket:
Ahh okay that is what was wrong, I should be more thorough from now on. Is there some log of what made a check fail? Or did you just notice the issue, if you don't mind my asking.

nimoalr:
> Is there some log of what made a check fail? Or did you just notice the issue, if you don't mind my asking.

You can have some clue of what is failing but not exactly why by looking at Travis CI build details by clicking here :

JiminyKroket:
Node.js: 8 ? I guess this would be the error line?

PR #257, opened 2020-02-20
CFX: Add missing drawing natives for RDR3

Disquse:
```cpp
void DRAW_LINE(float x1, float y1, float z1, float x2, float y2, float z2, int red, int green, int blue, int alpha);
void DRAW_POLY(float x1, float y1, float z1, float x2, float y2, float z2, float x3, float y3, float z3, int red, int green, int blue, int alpha);
```

Implemented in https://github.com/citizenfx/fivem/pull/323

### DIFF
--- a/CFX/DrawLine.md
+++ b/CFX/DrawLine.md
@@ -1,0 +1,24 @@
+---
+ns: CFX
+apiset: client
+game: rdr3
+---
+## DRAW_LINE
+
+```c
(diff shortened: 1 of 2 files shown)


stannum-cfx:
This is not the intended way to add RDR natives to RedM. This merge would break V natives. RDR natives are currently not supported in this repo.

PR #258, opened 2020-02-23
Update N_0x2d3b147afad49de0.md

utkuali:


GRAPHICS/N_0x2d3b147afad49de0.md
```diff
@@ -5,23 +5,42 @@ ns: GRAPHICS
 
 ```c
 // 0x2D3B147AFAD49DE0
-void _0x2D3B147AFAD49DE0(Any p0, Any p1, Any p2, Any p3, Any p4, Any p5, Any p6, Any p7, Any p8, Any p9, Any p10, Any p11);
+void _0x2D3B147AFAD49DE0(string, string, Any p2, Any p3, Any p4, Any p5, Any p6, Any p7, Any p8, Any p9, Any p10, Any p11);
```
Disquse: There's no such type `string`, use `char*` instead. Also you're missing an argument name, for example `char* p0`. According  to your description, you can name it as `minigameName` or so, if you not sure, use `p0`, `p1` and etc.

utkuali: I was trying to fit lua style

Disquse: Native definitions are done in C-style. You need to fix it if you want your PR to be merged.

utkuali: okay, thanks for letting me now

utkuali: @Disquse seems right?

Disquse: This is correct, but you can do more that this. Since you know meaning and/or types of some arguments, you could document it too. I will give you an example to let you know how it works. Feel free to edit this and push. Also I just noticed you're added `p12`, if you sure that `p12` is really exists, you should add it to the definition too.

```c
// 0x2D3B147AFAD49DE0
void _0x2D3B147AFAD49DE0(char* minigameName, char* minigamePart, float p2, float p3, Any p4, float p5, float p6, float p7, int red, int green, int blue, int alpha, Any p12);
```

Disquse: By a quick look around I can say that `p4` is a float too, and `p12` is a BOOL.

utkuali: Thank you for info, will use that as an example for the future pulls too.

utkuali: And by the way do I need to change * **p0** to minigameName and so on?

PR #274, opened 2020-04-09
cfx/mumble: fix file name typo

PichotM:
Rename MumbleSetVolumeOverrideByServer to MumbleSetVolumeOverrideByServerId


Frosty-Ice:
Ah, good catch. Don't know how I didn't notice.

PR #275, opened 2020-04-12
Create getVehicleInDirection.md

Hernior:


stannum-cfx:
Is this supposed to be a CFX native? If so it's in the wrong namespace.

Either way it's missing a hash

sgottelk:
Looks like part of the esx framework? https://esx-org.github.io/es_extended/client/functions/game/getvehicleindirection/

PR #276, opened 2020-04-14
added link to inputGroup docs.fivem.net

renaiku:
And updated the enum values to the matching ones instead of (unk)


alloc8or:
It is _not_ the input group.
0 definitely is ``PLAYER_CONTROL`` (from R* scripts) and 2 most likely is ``FRONTEND_CONTROL``.

sgottelk:
close?

PR #280, opened 2020-04-24
Fix dead URL

davwheat:


PAD/DisableControlAction.md
```diff
@@ -9,7 +9,7 @@ void DISABLE_CONTROL_ACTION(int inputGroup, int control, BOOL disable);
 ```
 
 ```
-control values and meaning: github.com/crosire/scripthookvdotnet/blob/dev_v3/source/scripting/Controls.cs  
+control values and meaning: https://github.com/crosire/scripthookvdotnet/blob/v3.0.3/source/scripting_v3/GTA/Control.cs
```
TomGrobbe, line 12: Some of those controls linked are wrong, why not link to the FiveM docs for a correct list of controls? https://docs.fivem.net/docs/game-references/controls/#controls

davwheat:
Because I didn't know it existed! 😅😅

PAD/DisableControlAction.md
sgottelk:
```suggestion
control values and meaning: https://docs.fivem.net/docs/game-references/controls/#controls
```

PR #282, opened 2020-04-27
Update GetEntityPopulationType.md

Quattro20v:
The above function description is from the client-side entity documentation. Was told on forum that both client-side and server-side return the same thing. Thought it might be helpful to have the description on the server-side also, for convenience


stannum-cfx:
Seems to me that yours is a duplicate @sgottelk 
anyway, CFX natives have been removed from this repo.

PR #301, opened 2020-05-21
retval description

interplar:
it returns false when in x,y,z position no collision 
so at first u need to load it and then GetGroundZFor_3dCoord


technetium-cfx:
@interplar Are you able to make the suggested changes?

technetium-cfx:
After testing, this appears to be incorrect. Just returns false when no ground coord was found. This happens if the specified Z coordinated is below ground level too. Current docs are correct.

PR #302, opened 2020-05-23
Update 0x317EBA71D7543F52 native to match format

polivilas:
Now it follows the format that all other natives do.


jaymo1011:
but... why?

PR #306, opened 2020-05-27
Update TaskVehicleDriveToCoord.md

arthur-rl:
Added some helpful maths


smallo92:
Could've just said the speed is in metres per second :P

technetium-cfx:
Agree with Smallo here. Closing.

PR #311, opened 2020-05-31
Update NetworkDoesEntityExistWithNetworkId.md

niekschoemaker:
Change entity to network id, since this could throw people off and break scripts if mixed up.

I did test this to make sure it expects a net id and this is the case.

### DIFF
--- a/NETWORK/NetworkDoesEntityExistWithNetworkId.md
+++ b/NETWORK/NetworkDoesEntityExistWithNetworkId.md
@@ -8,8 +8,8 @@ ns: NETWORK
 BOOL NETWORK_DOES_ENTITY_EXIST_WITH_NETWORK_ID(int netID);
 ```
 
-Gets wether the network ID exists and if this belongs to a game object.
+Gets if the network ID exists and if this belongs to a game object.
 ## Parameters
-* **netId**: 
+* **netID**: 
 
 ## Return value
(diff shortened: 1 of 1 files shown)


NETWORK/NetworkDoesEntityExistWithNetworkId.md
```diff
@@ -5,11 +5,11 @@ ns: NETWORK
 
 ```c
 // 0x18A47D074708FD68 0x1E2E3177
-BOOL NETWORK_DOES_ENTITY_EXIST_WITH_NETWORK_ID(Entity entity);
+BOOL NETWORK_DOES_ENTITY_EXIST_WITH_NETWORK_ID(int netID);
 ```
 
-
+Gets wether the network ID exists and if this belongs to a game object.
```
blattersturm, line 11: `wether`?

NETWORK/NetworkDoesEntityExistWithNetworkId.md
```diff
@@ -5,11 +5,11 @@ ns: NETWORK
...
 ## Parameters
-* **entity**: 
+* **netId**:
```
blattersturm, line 13: `netId` vs. `netID` in argument leads to build failure.

niekschoemaker: Didn't realise at the time that these files were being used for building.
Did realise it later on while trying to create a self made native.

I've fixed the issue

niekschoemaker:
I'm not sure about native names but seeing as what this native actually does the name also seems quite confusing since it actually checks if the Network Id Exists and an entity belongs to it.

technetium-cfx:
Already merged in 303.

PR #313, opened 2020-06-02
nativedb: SCRIPT_RACE

gottfriedleibniz:
Also contains a few fixes to natives in MISC/


blattersturm:
Renames seem to break with #309 being merged. Really, Git?

PR #317, opened 2020-06-15
datadict, dataarray, and miscellaneous

gottfriedleibniz:
The names for ``0x29280002282F1928`` and ``0x736D7AA1B750856B`` could likely be improved (e.g., DRAW_SPRITE_ONTO_POLY).


doctor-googler:
Hi @gottfriedleibniz ! Have you tried to use `0x29280002282F1928` and `0x736D7AA1B750856B` in scripts? 
I've tried to repeat the R*'s implementation of deadline and there are my results:
- using `0x29280002282F1928` I was able to render a colorized polygons with alpha. But it seems textures aren't applied for some reason. Texture dict was loaded before use, backside cooling was disabled and UVW mapping was repeated from original R* script.
- using `0x736D7AA1B750856B` I was able to render only black polygons and only alpha-per-vertex seems working. Also textures are not rendered for some reason

Do you have any thoughts on this? Thanks :)

PR #319, opened 2020-06-23
Update SetPedHeadOverlayColor.md

xdubx:


technetium-cfx:
Not all that useful. Perhaps we could display colors on our docs 🤔

PR #320, opened 2020-06-25
Update GetEntityMatrix.md

meta-hub:
Previous documentation seemed incorrect (at least with my experience in Lua).
`local forward,right,up,position = GetEntityMatrix(PlayerPedId())`


utkuali:
I can confirm this, tested it while trying to play custom surround sounds in html (nui) and it only works correctly with the return positions meta-hub corrected

PR #321, opened 2020-06-30
Update NewLoadSceneStart.md

Korioz:
i don't know much about C family so i tried my best :)


technetium-cfx (changes requested):
A bit vague, and not very consistent parameter naming compared to other natives. Try to speak from a third-person perspective, not the I-form. Also try to be sure of things, `probably` is definitely not very descriptive.

PR #327, opened 2020-07-08
Document few unknown natives

Matspyder51:


technetium-cfx:
Seems very vague. Any more information?

PR #331, opened 2020-07-17
Update IsVehicleAttachedToTrailer.md

Jill0:


technetium-cfx:
A bit too specifically odd example, we prefer examples that don't involve other logic than what the native is about.

PR #333, opened 2020-07-28
Update N_0x98215325a695e78a.md

skyrossm:
Allows you to enable/disable the mouse cursor.


technetium-cfx:
Good find. Perhaps suggest a nice name in another PR? 🙂

PR #335, opened 2020-08-03
Bool doc was false

X-rays5:
If i use true false or nil i get a error. but as soon as i change it to 1 and 0 it works just fine


technetium-cfx:
Could you explain what kind of error you are getting? This change seems counterintuitive, I'm not sure how to interpret it.

X-rays5:
seems to have been a bug on the newest server software it works fine

PR #340, opened 2020-08-13
Update N_0x76bba2cee66d47e9.md

KillerBoy001:


technetium-cfx (changes requested):
Please include an alias when renaming the native.

PR #350, opened 2020-09-02
Update SetEntityMaxSpeed.md

Sampleeeee:
Added a note that speed is in meters per second with examples of how to convert from different measurements of speed.

### DIFF
--- a/ENTITY/SetEntityMaxSpeed.md
+++ b/ENTITY/SetEntityMaxSpeed.md
@@ -8,6 +8,16 @@ ns: ENTITY
 void SET_ENTITY_MAX_SPEED(Entity entity, float speed);
 ```
 
+```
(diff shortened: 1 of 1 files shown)


ENTITY/SetEntityMaxSpeed.md
```diff
@@ -8,6 +8,16 @@ ns: ENTITY
 void SET_ENTITY_MAX_SPEED(Entity entity, float speed);
 ```
 
+```
```
technetium-cfx, line 11: No need to use codefences. This is only used for NativeDB imports.

technetium-cfx (changes requested):
To be honest, I think there are many more cases of conversion that people could simply google. Probably best to just mention that speed is in meters per second.

PR #352, opened 2020-09-04
update IsPedInVehicle.md

onvrb:
not working with given alternative

### DIFF
--- a/PED/IsPedInVehicle.md
+++ b/PED/IsPedInVehicle.md
@@ -19,3 +19,7 @@ If 'atGetIn' is false, the function will not return true until the ped is sittin
 * **atGetIn**: 
 
 ## Return value
+
+```
+Doesn't appear to work, use GET_VEHICLE_PED_IS_IN instead
+```


Sampleeeee:
This seems to work fine for me

```lua
function tobool(val)
	if val == nil or val == false or val == 0 or val == "0" or val == "false" then return false end
	return true
end

local lastCar
RegisterCommand('setlastcar', function(s, a) lastCar = GetVehiclePedIsIn(PlayerPedId()) end)

RegisterCommand("incar", function(s, a)
    a[1] = a[1] or "false"
    
    TriggerEvent("chat:addMessage", { 
        args = { 
            "Test", 
            "a[1] = "..a[1]..", "..tostring(IsPedInVehicle(
                PlayerPedId(), 
                lastCar, 
                tobool(a[1]))) 
        } 
    })
end)
```

Get in a car, do `/setlastcar` get in another car and do `/incar`, it returns false. If you're in the same car it returns true.

technetium-cfx:
Seems to work fine indeed. Closing.

PR #353, opened 2020-09-06
Update GetGroundZFor_3dCoord.md

gimicze:
Added a note about when the native returns true and a valid value.
Noticed this when trying to synchronize a fire on a OneSync Infinity server.


technetium-cfx:
Makes sense but does looping matter?
> When using this in a loop,

gimicze:
You're right, it doesn't. I wrote that to emphasize you could easily end up with indefinite loop when used incorrectly. Updated in last commit.

technetium-cfx:
Thanks for clarifying! Good PR.

PR #355, opened 2020-09-07
Modified SetVehicleLivery

Yimura:
Modified SetVehicleLivery to notify that it doesn't work and instead you should use SetVehicleMod

### DIFF
--- a/VEHICLE/SetVehicleLivery.md
+++ b/VEHICLE/SetVehicleLivery.md
@@ -8,6 +8,11 @@ ns: VEHICLE
 void SET_VEHICLE_LIVERY(Vehicle vehicle, int livery);
 ```
 
+```
+This native does nothing use:
+VEHICLE::SET_VEHICLE_MOD(veh, 48, index)
+```
+
 ## Parameters
 * **vehicle**: 
 * **livery**: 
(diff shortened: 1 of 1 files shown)


Frosty-Ice:
That's not true at all. There's vehicles such as the Windsor that use this native to set their livery.

Yimura:
Oh is it, I did use 1 small sample size of vehicles when applying Liveries to vehicles (10 or so) and had instant results with SetVehicleMod. Are there any other vehicles that work with SetVehicleLivery, do these work with SetVehicleMod?

Frosty-Ice:
The Sanchez is another. You can use GetVehicleLiveryCount to see if the vehicle has settable liveries via this method.

Yimura:
You appear to be correct, I tested it in-game and the Windsor does not use SetVehicleMod and GetNumVehMod does not work either for modIndex 48.

Yimura:
I've modified the markdown file accordingly to differentiate between vehicles that use `GetVehicleLiveryCount(veh)` and `GetNumVehicleMods(veh, 48)`

technetium-cfx:
I don't know, this makes very little sense to me.

PR #356, opened 2020-09-07
Made the DeleteEntity() API set shared

LedAndris:


blattersturm:
Wrong. This native has nothing to do with the server native which is in the citizenfx/fivem repo.

Shared natives by definition should not exist in this repo whatsoever.

LedAndris:
Smallo lightened me up

PR #363, opened 2020-10-04
fix #168

### LINKED ISSUE #168: Wrong HASH for _SET_OBJECT_COLOUR
_SET_OBJECT_COLOUR has 0x3B2FD68DB5F8331C hash in docs which is wrong. The correct one would be 0x5F048334B4A4E774

gottfriedleibniz:
fix #168


technetium-cfx:
Description moved to a different native? Looks okay, just making sure this was intended 😛

gottfriedleibniz:
Yeah, `0x3B2FD68DB5F8331C` is a 757 native that looked to be incorrectly mapped. You can also reference alloc8or's database as he initially pointed it out (if I remember correctly).

PR #365, opened 2020-10-06
Update GetNameOfZone.md

DokaDoka:
replaced a duplicate entry of "La Puerta" with "Puerto Del Sol" at DELSOL


technetium-cfx:
Seems accurate, although `GetLabelText('DELSOL')` returns `La Puerta`. Not sure what should be leading here.

technetium-cfx:
@DokaDoka Are you able to provide clarification?

PR #366, opened 2020-10-08
Update IsVehicleWindowIntact native documentation

xotikorukx:
This update adds a line of documentation to better describe a failure case of this native.


technetium-cfx:
That's okay, there may be more natives that do this. Should be within the scope of "expectations" when coding something that may go out of range.

PR #369, opened 2020-10-15
CreateVehicle Update

GianmarcoAquilano:
CreateVehicle: netMissionEntity --> thisScriptCheck


blattersturm:
no? what makes you think it's "thisScriptCheck" which means nothing? it's literally network mission entity.

PR #372, opened 2020-10-19
Update SetPedPopulationBudget.md

Vivi4n:
Adding description to Ped budget as tested, as is not documented.


STREAMING/SetPedPopulationBudget.md
```diff
@@ -8,6 +8,10 @@ ns: STREAMING
 void SET_PED_POPULATION_BUDGET(int p0);
 ```
 
+```
```
blattersturm, line 11: Manual descriptions shouldn't be written in \`\`\`. This is only used for NativeDB imports.

PR #374, opened 2020-10-24
Update StopFireInRange.md

JamesUK-Developer:


technetium-cfx:
We should not treat descriptions as a section for discussions. Please keep descriptions accurate and clear 😄 
Documented examples can go into a dedicated examples section, as shown [here](https://github.com/citizenfx/natives/blob/master/.github/CONTRIBUTING.md#example-documentation).

PR #379, opened 2020-10-30
Update TaskLeaveAnyVehicle.md

Sampleeeee:
I added an example for the use of the native and renamed p2 to flags

Wasn't sure if I should just copy and paste flags from TaskLeaveVehicle or just say to look at TaskLeaveVehicle. Assumed it would be better to make them refer to TaskLeaveVehicle in case any flags are added to that list

### DIFF
--- a/TASK/TaskLeaveAnyVehicle.md
+++ b/TASK/TaskLeaveAnyVehicle.md
@@ -8,9 +8,7 @@ ns: TASK
 void TASK_LEAVE_ANY_VEHICLE(Ped ped, int p1, int flags);
 ```
 
-```
-Flags seem to be the same flags used in TaskLeaveVehicle
-```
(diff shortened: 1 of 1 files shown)


technetium-cfx:
Nice find. Do you have any example code that demonstrates they are the same as `TaskLeaveVehicle`?
Few notes:
- Don't codefence descriptions with ```. That's only used for NativeDB imports.
- You can link to `TaskLeaveVehicle` using `[TaskLeaveVehicle](#0xHASH)`
- `seem to be` sounds a bit unsure, is `are` better?

technetium-cfx (changes requested):
Please review suggested changes 🙂

Sampleeeee:
Is that what you meant by removing the code fence from the description?

PR #383, opened 2020-11-03
Update GetZoneFromNameId.md

Celeron64:


technetium-cfx:
What made you add this item? With some quick testing it appears that the entire list in this description makes no sense at all. Furthermore, the item you added does not seem to appear in `popzone.ipl` either.

PR #385, opened 2020-11-07
Update SetHighFallTask.md

LedAndris:
I tested a lot of params for p1 and p2 but wasn't able to find out any serious consistency between the flags so this just gives a hand to developers trying with ragdolls, but the second parameter is for sure the duration I'm certain about that.

### DIFF
--- a/TASK/SetHighFallTask.md
+++ b/TASK/SetHighFallTask.md
@@ -5,26 +5,27 @@ ns: TASK
 
 ```c
 // 0x8C825BDC7741D37C 0xBBB26172
-void SET_HIGH_FALL_TASK(Ped ped, int p1, Any p2, Any p3);
+void SET_HIGH_FALL_TASK(Ped ped, Any duration, Any p2, Any p3);
 ```
 
 ```
-Makes the ped ragdoll like when falling from a great heigh.
-Playing with p1 and p2 changes the animation's style and behavior.
+p1 seems to do something with the duration of the animation
+Makes the ped ragdoll like when falling from a great height.
+Playing with p2 and p3 changes the animation's style and behavior.
 In most cases the ped moves backwards then fall on the ground or collapse with a very small change to it's position.
-If p1 or p2 is 1 the ped will move backwards  
-On p1: 0, p2: 0 the ped will violently move backwards before falling.
-On p1: 1, p2: 1 the ped will violently move to a random direction.
-On p1: 2, p2: 2 the ped just collapses like a bag of potato.
+If p2 or p3 is 1 the ped will move backwards 
+On p2: 0, p3: 0 the ped will violently move backwards before falling.
+On p2: 1, p3: 1 the ped will violently move to a random direction.
+On p2: 2, p3: 2 the ped just collapses like a bag of potato.
 Good combinations tried:
-p1: 100 p2: 1000
-p1: 100 p2: 1
+p2: 100 p3: 1000
+p2: 100 p3: 1
 Even numbers work best
 ```
 
 ## Parameters
 * **ped**: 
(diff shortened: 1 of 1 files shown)


TASK/SetHighFallTask.md
```diff
@@ -5,16 +5,26 @@ ns: TASK
 
 ```c
 // 0x8C825BDC7741D37C 0xBBB26172
-void SET_HIGH_FALL_TASK(Ped ped, Any p1, Any p2, Any p3);
+void SET_HIGH_FALL_TASK(Ped ped, int p1, Any p2, Any p3);
 ```
 
 ```
```
technetium-cfx, line 11: Please remove the codefence, this is only used by NativeDB imports.

TASK/SetHighFallTask.md
```diff
@@ -5,16 +5,26 @@ ns: TASK
...
-Makes the ped ragdoll like when falling from a great height  
+Makes the ped ragdoll like when falling from a great heigh.
```
technetium-cfx, line 12: `heigh`? 😛

TASK/SetHighFallTask.md
```diff
@@ -5,16 +5,26 @@ ns: TASK
...
+On p1: 2, p2: 2 the ped just collapses like a bag of potato.
+Good combinations tried:
+p1: 100 p2: 1000
+p1: 100 p2: 1
+Even numbers work best
 ```
 
 ## Parameters
 * **ped**: 
+* **duration**:
```
technetium-cfx, line 27: Parameters here must match parameter names in the function signature.

TASK/SetHighFallTask.md
```diff
@@ -5,16 +5,26 @@ ns: TASK
...
-void SET_HIGH_FALL_TASK(Ped ped, Any p1, Any p2, Any p3);
+void SET_HIGH_FALL_TASK(Ped ped, int p1, Any p2, Any p3);
 ```
 
 ```
-Makes the ped ragdoll like when falling from a great height  
+Makes the ped ragdoll like when falling from a great heigh.
+Playing with p1 and p2 changes the animation's style and behavior.
+In most cases the ped moves backwards then fall on the ground or collapse with a very small change to it's position.
+If p1 or p2 is 1 the ped will move backwards
```
technetium-cfx, line 15: You mention p1 is duration, but this does not suggest that. Could you elaborate on this?

TASK/SetHighFallTask.md
```diff
@@ -5,16 +5,26 @@ ns: TASK
...
+p1: 100 p2: 1
+Even numbers work best
 ```
 
 ## Parameters
 * **ped**: 
+* **duration**: 
 * **p1**: 
 * **p2**: 
-* **p3**:
```
technetium-cfx, line 19: p3 should still be the last param. Index is not to be changed.

technetium-cfx (changes requested):
Left some feedback. Useful information to have, but should be a bit clearer 🙂

TASK/SetHighFallTask.md
LedAndris: Oh I see, thank you (that's why I used p1 and p2)

LedAndris:
Will keep these informations in mind, thank you for your feedback.

technetium-cfx:
You requested a review, but no changes have been made yet 😛

LedAndris:
Thats strange, I'm quite sure I made the requested changes, but this is probably ma fault so let me retry

technetium-cfx:
Still nothing!

LedAndris:
Left a commit, I suck with this part of Github, sorry about this

PR #392, opened 2020-11-26
Update CreateCam making note of breaking bug

Fires1:


Fires1:
This may not be in the regular spirit of how these native references work, but considering the only reference to a solution is on this forum post. It may be in the best interest to include this as there are no indicators (besides the linked post) on how to resolve this issue.

https://forum.cfx.re/t/c-updates-issue-native-methods-no-longer-running-after-the-last-update-bug-update/134394

Fires1:
I stand corrected on one instance, actually, I am yet to test if you can just trigger the client event from the client and avoid the server entirely...will update with a following commit if true...

PR #393, opened 2020-11-26
Update CreateCam.md

Fires1:


technetium-cfx:
What do you mean by the "console command manager thread"? Could you provide some example code to demonstrate in what way this note is necessary?

technetium-cfx:
As a matter of fact, the issue you were experiencing has just been fixed: https://github.com/citizenfx/fivem/commit/c1ef032f404644dc14fcd0a049a6a31174fe05f3

If you encounter other issues with functions, please report them as a bug. That way we can fix them 🙂

PR #394, opened 2020-12-03
Update TRIGGER_SCREENBLUR_FADE_IN

ProgrammingHider:
I added context on what float transitionTime does on TRIGGER_SCREENBLUR_FADE_IN


AvarianKnight (changes requested):
Shouldn't the description be beside the parameter?

PR #395, opened 2020-12-04
Update GetClosestVehicle.md

ryanking44:
Added note about flag 127.  Seemed to work to return police cars for me


technetium-cfx:
"for me" isn't very helpful. Have you thoroughly tested this and do you have a test case to verify your findings?

FlawwsX:
After testing this, if the flag is set to `127` it explicitly returns police vehicles, does not appear to return anything else.

**I tested using the following code:**
```
CreateThread(function()
    while true do 
        local plyPos = GetEntityCoords(PlayerPedId())
        local veh = GetClosestVehicle(plyPos.x, plyPos.y, plyPos.z, 10.0, 0, 127)
        print(GetDisplayNameFromVehicleModel(GetEntityModel(veh)), veh)
        Wait(1000)
    end
end)
```

PR #400, opened 2020-12-07
Update IsVehicleWindowIntact.md

spoty123:
Description of this native, especially to show that you could use it to check if a window is rolled down.
Tested with this code:

```lua
local ped = PlayerPedId()
local vehicle = GetVehiclePedIsIn(ped, false)
local isIntact = IsVehicleWindowIntact(vehicle, 0)
print(isIntact) -- 1
RollDownWindow(vehicle, 0)
isIntact = IsVehicleWindowIntact(vehicle, 0)
print(isIntact) -- false
```

### DIFF
--- a/VEHICLE/IsVehicleWindowIntact.md
+++ b/VEHICLE/IsVehicleWindowIntact.md
@@ -12,6 +12,14 @@ BOOL IS_VEHICLE_WINDOW_INTACT(Vehicle vehicle, int windowIndex);
 This will return false if the window is broken, or rolled down.
 ```
 
+```
+windowIndex:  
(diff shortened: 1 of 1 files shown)


technetium-cfx (changes requested):
Please review the [contributing guidelines](https://github.com/citizenfx/natives/blob/master/.github/CONTRIBUTING.md).

- Descriptions should not be in code blocks
- Descriptions should come before parameters

Additionally, a vehicle may have more than 4 windows. Probably best if this is explained.

spoty123:
Pushed a new commit that changes the structure and adds info as you requested.

PR #401, opened 2020-12-20
Update TaskOpenVehicleDoor.md

gimicze:
I was playing around with this native when writing a script of mine and found out, how it works (~).

The following code tasks the ped to open the specified door according to[ SET_VEHICLE_DOOR_OPEN's](https://runtime.fivem.net/doc/natives/?_0x7C65DAC73C35C862) door indexes
```
if nearestDoorIndex < 4 then -- doorIndex is index according to table in SET_VEHICLE_DOOR_OPEN reference
    TaskOpenVehicleDoor(playerPed, nearestVehicle, 1.0, (nearestDoorIndex - 1), 1.0)
end
````

### DIFF
--- a/TASK/TaskOpenVehicleDoor.md
+++ b/TASK/TaskOpenVehicleDoor.md
@@ -16,6 +16,8 @@ void TASK_OPEN_VEHICLE_DOOR(Ped ped, Vehicle vehicle, int timeOut, int seat, flo
 * **speed**:
 
 ```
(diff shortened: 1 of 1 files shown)


gimicze:
Found out that the native is very similar to [TASK_ENTER_VEHICLE](https://runtime.fivem.net/doc/natives/?_0xC20E50AA46D09CA8). It behaves quite the same, except the ped never gets into the vehicle, which means for the front and rear doors, the ped would get into the car through the nearest door and then change seats. Since this native never allows the ped to get in, the ped follows the nearest door of the type (front / rear).

PR #402, opened 2020-12-26
Update StartGpsMultiRoute.md

TheRealFloatDev:
Added C# example


Disquse:
C# example with Lua-styled commentaries? Should I mention that this syntax is invalid and it won't compile?
Replace `--` with `//`.

HUD/StartGpsMultiRoute.md
```diff
@@ -42,3 +42,20 @@ AddPointToGpsMultiRoute(1638.2155761719, -2344.685546875, 95.610511779785)
 -- Set the route to render
 SetGpsMultiRouteRender(true)
 ```
+```C#
```
Disquse, line 45: Also C# code blocks should be notated with `css` tag.

PR #407, opened 2021-01-07
Clarifies prerequisites to displaying player blip labels on bigmap

tomezpl:
If you try to display player blip labels on the expanded map using [0x82CEDC33687E1F50](https://runtime.fivem.net/doc/natives/?_0x82CEDC33687E1F50), you need to call [SetBlipCategory](https://runtime.fivem.net/doc/natives/?_0x234CDD44D996FD9A) with index 7 as otherwise the labels won't display. Here's a screenshot of the expanded minimap with a blip of category 2: 
image

Here's category 7 (working as expected) 
image

This PR just clarifies that requirement.

### DIFF
--- a/HUD/N_0x82cedc33687e1f50.md
+++ b/HUD/N_0x82cedc33687e1f50.md
@@ -11,6 +11,8 @@ Toggles whether or not name labels are shown on the expanded minimap next to pla
 Doesn't need to be called every frame.  
 Preview: https://i.imgur.com/DfqKWfJ.png
 
+Make sure to call [SetBlipCategory](https://runtime.fivem.net/doc/natives/?_0x234CDD44D996FD9A) with index 7 for this to work on the desired blip.
+
 
 ## Parameters
 * **p0**: the toggle boolean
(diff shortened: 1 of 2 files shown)


gottfriedleibniz:
0x2F28F0A6 = DISPLAY_PLAYER_NAME_TAGS_ON_BLIPS ; if you would like to actually name the native as well.

tomezpl:
> 
> 
> 0x2F28F0A6 = DISPLAY_PLAYER_NAME_TAGS_ON_BLIPS ; if you would like to actually name the native as well.

Yup, added that

HUD/N_0x82cedc33687e1f50.md
```diff
@@ -1,16 +1,18 @@
 ---
 ns: HUD
 ---
-## _0x82CEDC33687E1F50
```
technetium-cfx, line 4: When renaming natives, please rename the file and include an alias as well. An example can be found in [here](https://github.com/citizenfx/natives/blob/master/.github/CONTRIBUTING.md#guidelines).

technetium-cfx (changes requested):

PR #408, opened 2021-01-08
Added parameter description

Veteranfighter:
description added for the following parameters: p2, x, y, z, cargobob, vehicle


technetium-cfx (changes requested):
'Some type of offset' isn't all too descriptive, would you happen to have a better, more accurate description for this?

PR #409, opened 2021-01-14
Added flag doc for SetTaskVehicleChaseBehaviorFlag

Rubylium:


technetium-cfx (changes requested):
[Remember](https://github.com/citizenfx/natives/blob/master/.github/CONTRIBUTING.md#guidelines), descriptions come under the function definition, without the code fence 😄

Rubylium:
Better now ? :)

technetium-cfx:
Definitely! Would be nice if it was ordered, but we can do that before merging if still necessary. Would you happen to have any code to validate this change for us?

PR #410, opened 2021-01-22
Update NetworkSessionEnter.md

PatchByte:
Changed the return value to bool


NETWORK/NetworkSessionEnter.md
```diff
@@ -29,3 +29,4 @@ Global_1315318 = 0 or Global_1315323 = 9 or 12 or (Global_1312629 = 0 && Global_
 * **p5**: 
 
 ## Return value
+BOOL
```
technetium-cfx, line 32: The description for the return value is not the type; rather, an explanation of what data is returned.

technetium-cfx (changes requested):

PR #418, opened 2021-02-04
Update NetworkSessionSetMatchmakingGroup.md

PatchByte:
Added the pointer + offset and renamed the name!


technetium-cfx:
Please review our [contribution guidelines](https://github.com/citizenfx/natives/blob/9b42ebaa8de86f073300044d47d04bc91032217e/.github/CONTRIBUTING.md); this change requires an alias and should rename the file.

Would you have more accurate information on this change as well? Why did you rename it? What to do with the pointer? Most users won't know what to do with a pointer, and it doesn't explain a whole lot either.

PR #421, opened 2021-02-07
Updated description of native

NylanderrDK:
The pastebin doesn't exist anymore, so it's now deleted.


technetium-cfx:
Would be useful to update it with a new list somewhen.

PR #425, opened 2021-02-11
Rename N_0x9e4cfff989258472.md to InvalidateVehicleIdleCam.md

ElPumpo:


CAM/InvalidateVehicleIdleCam.md
```diff
@@ -0,0 +1,23 @@
+---
+ns: CAM
+aliases: ["_0x9E4CFFF989258472"]
+---
+## INVALIDATE_VEHICLE_IDLE_CAM
+
+```c
+// 0x9E4CFFF989258472 0x65DDE8AF
+void INVALIDATE_VEHICLE_IDLE_CAM();
```
Disquse, line 9: Seems that this name doesn't fit joaat hash (i.e. that's not a real name of this native). In this case you should add a `_` prefix to this name.

ElPumpo: Disquise long time ago, yea I was unsure if it needed a _ or not.

CAM/InvalidateVehicleIdleCam.md
```diff
@@ -5,7 +5,7 @@ aliases: ["_0x9E4CFFF989258472"]
 ## INVALIDATE_VEHICLE_IDLE_CAM
 
 ```c
-// 0x9E4CFFF989258472 0x65DDE8AF
+// _0x9E4CFFF989258472 0x65DDE8AF
 void INVALIDATE_VEHICLE_IDLE_CAM();
```
Disquse, line 9: Sorry for me being not enough clear :(
I mean native's name should be prefixed, not hash.

ElPumpo: Ok I reverted both _ and added _ to the raw native name instead. So if the raw native name doesn't match the hash then we add a underscore to it?

technetium-cfx: Correct. You can review the process in our [contribution guidelines](https://github.com/citizenfx/natives/blob/master/.github/CONTRIBUTING.md) as well. (Which you should have read 😛)

ElPumpo: Okok well I pushed a new commit fixing it

CAM/InvalidateVehicleIdleCam.md
```diff
@@ -0,0 +1,23 @@
+---
+ns: CAM
+aliases: ["0x9E4CFFF989258472"]
+---
+## _INVALIDATE_VEHICLE_IDLE_CAM
+
+```c
+// 0x9E4CFFF989258472 0x65DDE8AF
+void INVALIDATE_VEHICLE_IDLE_CAM();
```
Disquse, line 9: This line needs prefix too 😅
See other natives for example: https://github.com/citizenfx/natives/blob/master/LOADINGSCREEN/LoadingscreenSetLoadFreemode.md

ElPumpo: Oh no I missed it completly! Thought there only was one

ElPumpo:
Ok all fuck ups should be resolved now

PR #430, opened 2021-03-06
Updatet SetPlayerHealthRechargeLimit

Korgron:
Added further information to provide more understanding of how to use this native properly

### DIFF
--- a/PLAYER/SetPlayerHealthRechargeLimit.md
+++ b/PLAYER/SetPlayerHealthRechargeLimit.md
@@ -9,10 +9,11 @@ aliases: ["0xC388A0F065F5BC34"]
 void _SET_PLAYER_HEALTH_RECHARGE_LIMIT(Player player, float limit);
 ```
 
+Description of the native
+
 ## Parameters
 - **player**: PlayerId()
 - **limit**: float number between 0.1 and 0.9
 
-## Important
-It seems that it only accepts numbers between 0.1 and 0.9. Numbers like 1.0 and upper doesn't work
 
+It seems that it only accepts numbers between 0.1 and 0.9. Numbers like 1.0 and upper doesn't work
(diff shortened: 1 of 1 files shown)


PLAYER/SetPlayerHealthRechargeLimit.md
```diff
@@ -9,8 +9,10 @@ aliases: ["0xC388A0F065F5BC34"]
...
 ```
 
-
 ## Parameters
-* **player**: 
-* **limit**: 
+- **player**: PlayerId()
+- **limit**: float number between 0.1 and 0.9
+
+## Important
```
technetium-cfx, line 16: Documentation does not support additional headings. Please include this as description.

technetium-cfx (changes requested):
Thanks for the contribution! Do have any code for us to verify this change?

Korgron:
## Example code as client script in Javascript
```JS
Wait = (ms) => new Promise(resolve => setTimeout(resolve, ms))

RegisterCommand("setHealth", async () => {
    let ped = PlayerPedId()
    SetEntityMaxHealth(ped, 500) //Set Health over the default health amount to test, if the health will be recharged to 90% (0.9)
    SetPlayerHealthRechargeMultiplier(PlayerId(), 20.0) //Speed up recharging health
    SetPlayerHealthRechargeLimit(PlayerId(), 0.9) //Set rechargelimit to 0.9 (90%)
})
//Our health should be recharged to 450HP

setTick(async () => {       //create an async loop. Similar to Citizen.CreateThread(function() end)
    let ped = PlayerPedId()
    let health = GetEntityHealth(ped)
    console.log(health)  //log the amount of health to the console
    await Wait(1000)    // Wait every one second
})
```

Disquse:
Why do you use `setTick` with 1000 delay when you can just create an interval?

Korgron:
Well, I haven't used `setInterval()` yet, but I will give this a try

Korgron (commented):
I have removed the ##Important headline and added "Description of the native"

technetium-cfx:
Huh, that change makes no sense. Why'd you add a bare "Description of the native"? The point of the example in the contribution guidelines is to highlight where the description goes, lol 😛

Korgron:
Oh, lol, okay, then I just remove it 😄

PR #449, opened 2021-04-06
docs: update SetPedKeepTask native

ctsmax:


PED/SetPedKeepTask.md
```diff
@@ -7,9 +7,14 @@ ns: PED
...
 
 ## Parameters
-* **ped**: 
-* **toggle**: 
+* **ped**: The given ped
+* **toggle**: Whether or not the ped will keep his task
 
+## Examples
+```lua
+local ped = CreatePed("PED_TYPE_CIVMALE", "cs_solomon", -1027.192, -2730.865, 12.75, 110.0, false, true)
```
Disquse, line 18: This code is invalid. First arg of `CREATE_PED` is integer, not string. And ped won't be created because model is not loaded.

ctsmax: I didn't think it was necessary to put all the code before the ped creation (including the model loading etc..)
Also the ```CREATE_PED``` native can be used like this.

ctsmax:
This is what you wanted I guess.

PR #458, opened 2021-04-29
Patch 6

niekschoemaker:
Added documentation of types for Cos and Sin native.

Usefull to avoid mistakes when converting this to native lua code (since math.cos and math.sin is probably a more performant option)

### DIFF
--- a/SYSTEM/Cos.md
+++ b/SYSTEM/Cos.md
@@ -10,7 +10,6 @@ float COS(float value);
 
 ```
 Returns the cosine of the given number.
-
 ```
 
 ## Parameters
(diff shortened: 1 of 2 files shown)


blattersturm:
Patch 6 :o

SYSTEM/Cos.md
```diff
@@ -8,8 +8,22 @@ ns: SYSTEM
 float COS(float value);
 ```
 
+```
```
technetium-cfx, line 11: Descriptions should not be [codefenced](https://github.com/citizenfx/natives/blob/master/.github/CONTRIBUTING.md#descriptions) :) Will fix, but a good reminder for the future.

PR #459, opened 2021-05-04
CONVERT_POSIX_TIME C# struct edit

BennoArnautovic:
ConvertPosixTime returns only year, added functions if someone else needs it

### DIFF
--- a/NETWORK/ConvertPosixTime.md
+++ b/NETWORK/ConvertPosixTime.md
@@ -21,6 +21,28 @@ struct date_time
     alignas(8) int second;  
 };
 ```
+```lua
+In Lua, this wont work, since it returns C# struct, these are functions that you can use to get time as os.date server-side is returning:
+local function GetIntFromBlob(b, s, o)
+	r = 0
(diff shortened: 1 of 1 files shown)


gottfriedleibniz:
For Lua 5.3 this is very much a hack. 

If 'blob' is less-than-or-equal to 40 characters (LUAI_MAXSHORTLEN) than the string will be internalized (https://en.wikipedia.org/wiki/String_interning) and manipulated externally. For strings this size and greater, the 'hash' (https://github.com/citizenfx/fivem/blob/master/code/vendor/lua/src/lobject.h#L369) of the string object will be computed and cached whenever/if-ever the string is inserted into a table. Both cases have the potential to leave Lua in a technically invalid state.

This native repository is not the place to endorse these hacks. Although I have no issues with powerusers being aware of them. In addition, improved support for struct parsing for the-best-la... Lua may come in the near future.

technetium-cfx:
Might be better for this to live on the main docs somewhere indeed.

PR #461, opened 2021-05-05
Update SetVehicleFixed.md

chipjpg:
Added potential issue


technetium-cfx:
The documentation is not a place for research discussion. If you encounter a bug, it's best to report it on the forums so we can actually provide a fix.

PR #462, opened 2021-05-06
Create CreateMissionTrainByName.md

chipjpg:


gottfriedleibniz:
This repository only maintains natives from PC/NG and many of the 32-bit Jenkins you do see are from mapping the jenkins hashes of early console builds to PC. If you have a significant number of (early-) console-specific natives documented, we could probably use them to ensure no native has been missed (or incorrectly mapped).

Otherwise, this repository isn't the place :(

PR #467, opened 2021-05-19
added note about parameter types

Whoeza:
The function also accepts ped names passed as strings, not only model hashes.


Whoeza:
Hey I had wrote a more detailed commit, I don't see my changes being applied. I don't understand where they went..

PR #469, opened 2021-05-19
added note about parameter types and an example

Whoeza:
The function also accepts ped names passed as strings, not only model hashes.

### DIFF
--- a/STREAMING/RequestModel.md
+++ b/STREAMING/RequestModel.md
@@ -11,7 +11,22 @@ void REQUEST_MODEL(Hash model);
 ```
 Request a model to be loaded into memory.
 ```
+**Note**: the function **also** accepts a model passed by ped name. (string)
+
+e.g.: RequestModel('ig_abigail')
+
+Same as: RequestModel(0x400AEC41)
+
 
 ## Parameters
 * **model**: 
 
+## Examples
+```lua
+local enemyInfo = {model ="csb_ramp_marine",
+  coords = vector3(playerSpawnCoords.x + 4.5,        
+  playerSpawnCoords.y + 1.5,
+  playerSpawnCoords.z),
+  heading = 180.0}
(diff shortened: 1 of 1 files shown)


STREAMING/RequestModel.md
```diff
@@ -11,7 +11,22 @@ void REQUEST_MODEL(Hash model);
 ```
 Request a model to be loaded into memory.
 ```
+**Note**: the function **also** accepts a model passed by ped name. (string)
```
blattersturm, line 14: This is not specific to 'the function', and is incorrect: the Lua/JS script runtimes will hash any string passed to a 'Hash' type. Your note would not apply to C#, for example.

STREAMING/RequestModel.md
```diff
@@ -11,7 +11,22 @@ void REQUEST_MODEL(Hash model);
...
+
+Same as: RequestModel(0x400AEC41)
+
 
 ## Parameters
 * **model**: 
 
+## Examples
+```lua
+local enemyInfo = {model ="csb_ramp_marine",
```
blattersturm, line 26: A bit of a useless example as this native is usually not used in isolation (e.g. a model would usually get awaited later on).

STREAMING/RequestModel.md
Whoeza: Ah. Didn't know that. JS and Lua will call GetHashKey by themselves? Or do they use another hash function that has the same effect of the native GetHashKey? Have you tried it in C#?

STREAMING/RequestModel.md
technetium-cfx: That code is also not very useful; you probably meant `while not HasModelLoaded(enemyInfo.model) do`. Anyhow, this change doesn't add too much extra information. The main docs repo would be a better place to document Lua/JS specific support.

PR #477, opened 2021-06-06
Added lua examples to PlaySoundFrontend.md

CritteRo:
Added some frontend sound examples that I use in my scripts and know that they work.


technetium-cfx:
This is a bit overcomplicated for what it does.

PR #478, opened 2021-06-08
Update CreateCam.md

BJDubb:
Added note when creating cameras. Found this out the hard way lol

Validation:

Create many camera's in a while loop and set them as active. The game will not create new cameras after a certain amount have been created. The number appears to differ depending on unknown variables. 

In order to be able to create new cameras, you must delete some to empty the "camera buffer"

### DIFF
--- a/CAM/CreateCam.md
+++ b/CAM/CreateCam.md
@@ -13,7 +13,14 @@ Cam CREATE_CAM(char* camName, BOOL p1);
 "DEFAULT_ANIMATED_CAMERA"  
 "DEFAULT_SPLINE_CAMERA"  
 "DEFAULT_SCRIPTED_FLY_CAMERA"  
-"TIMED_SPLINE_CAMERA"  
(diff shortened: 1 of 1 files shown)


technetium-cfx:
We generally look for accurate, non-confusing information when [writing](https://github.com/citizenfx/natives/blob/master/.github/CONTRIBUTING.md#descriptions) documentation. There is probably a limit of cameras the game can have, but this doesn't explain that limit.

BJDubb:
> We generally look for accurate, non-confusing information when [writing](https://github.com/citizenfx/natives/blob/master/.github/CONTRIBUTING.md#descriptions) documentation. There is probably a limit of cameras the game can have, but this doesn't explain that limit. It would be better to include accurate information on that. 🙂

I completely understand. I just feel like a little side note may be helpful to someone who is not aware of this. It certainly would have helped me in the 30-40 mins where I was pulling my hair out lol.

PR #481, opened 2021-06-12
Update parameter range for SetControlNormal.md

TheGreenkey:
This expands the allowed value range for the `amount` parameter to include negative values. For example, the `INPUT_VEH_MOVE_UD` control can accept `-1.0` for the other direction.


gottfriedleibniz:
This is based off a thirty second glance at the native.

The native accepts any fpvalue and does no input sanitization. It is possible its name is actually `_SET_CONTROL_UNBOUND_NORMAL`. The decompiled scripts even use 3.f for `INPUT_SCRIPTED_FLY_ZUP` and `INPUT_SCRIPTED_FLY_ZDOWN`.

You could replace the comment "A value from -1.0 to 1.0" with "An unbounded normal value". Leaving the comment open ended enough so if the control system is ever formally documented, a link in the comment may be included.

TheGreenkey:
I changed the parameter description for now. Like you said, if someone properly documents the control system in the future, the name could then be changed too.

PR #482, opened 2021-06-12
AttachContainerToHandlerFrame

MRV6:
Snippet : https://pastebin.com/NSMc1xZp


gottfriedleibniz:
`Entity container` > `Object container`

Should include a comment to note that the native is hard-coded to only work with `prop_contr_03b_ld` objects. I tried brute-forcing the hash to no avail. ):

technetium-cfx:
The snippet provided might even serve as a good example, perhaps worth adding?

MRV6:
Should I add the whole code in the markdown file or do you want to host it somewhere ?

PR #495, opened 2021-07-02
Update PointCamAtEntity.md

The-Neco:
Named unknown parameters for PointCamAtEntity.

P2, P3 and P4 are the XYZ offset of the entity.

Code to test:
Camera points at the marker, therefor P2, P3 and P4 are entity offsets.

```lua
local offset = vec3(1.0, 0.0, 0.0)
local cam = CreateCameraWithParams("DEFAULT_SCRIPTED_CAMERA", GetEntityCoords(PlayerPedId()), 0.0, 0.0, 0.0, 90.0, true, 2)
PointCamAtEntity(cam, PlayerPedId(), offset, true)
RenderScriptCams(true, false, 0, true, true)

  while true do
    Citizen.Wait(0)
    local coords = GetOffsetFromEntityInWorldCoords(PlayerPedId(), offset)
    DrawMarker(0, coords, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 1.0, 1.0, 0.5, 255, 255, 255, 255)
  end
end)
```

### DIFF
--- a/CAM/PointCamAtEntity.md
+++ b/CAM/PointCamAtEntity.md
@@ -8,15 +8,21 @@ ns: CAM
 void POINT_CAM_AT_ENTITY(Cam cam, Entity entity, float p2, float p3, float p4, BOOL p5);
 ```
 
-```
-p5 always seems to be 1 i.e TRUE  
-```
+Points the camera at the specified entity.
+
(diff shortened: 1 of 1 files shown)


FreakOutModz (commented):
Perfect way to improve a native.

PR #496, opened 2021-07-03
Update and rename SetCheckpointScale.md to SetCheckpointIconHeight.md

TheGreenkey:


GRAPHICS/SetCheckpointIconHeight.md
```diff
@@ -0,0 +1,22 @@
...
+
+This multiplies the height of the icon inside a checkpoint with the default height of about 2 units above the checkpoint's coordinates.
+
+## Parameters
+* **checkpoint**: The checkpoint the multiplier should be applied to.
+* **height_multiplier**: The multiplier for the icon height above the checkpoint's coordinates. Default is 1.0
+
+## Examples
+```lua
+local checkpoint = CreateCheckpoint(...)
```
stannum-cfx, line 20: This example won't work when copy/pasting, but it's not a major issue, try to make a copy/pastable example in the future.

TheGreenkey: Alright, wasn't aware that they should be copy/pastable without modification. Will keep that in mind!

PR #499, opened 2021-07-05
Update HasEntityClearLosToEntity.md

niekschoemaker:


gottfriedleibniz:
Given the updated parameter descriptions, you could remove everything in the `flags is always 17 in the scripts (object | map).` codeblock.

niekschoemaker:
Makes sense, I changed it now

PR #504, opened 2021-07-09
Update N_0x53f4892d18ec90a4.md

crocodildos:


MOBILE/CellCamSetDistance.md
```diff
@@ -0,0 +1,18 @@
+---
+ns: MOBILE
+aliases: ["0x53F4892D18EC90A4"]
+---
+## CELL_CAM_SET_DISTANCE
+
+```c
+// 0x53F4892D18EC90A4
+void CELL_CAM_SET_DISTANCE(float p0);
```
Disquse, line 9: If this name is not verified (I can't see joaat hash here, so it's not), there should be a `_` prefix. Same for line 5.

MOBILE/CellCamSetDistance.md
```diff
@@ -0,0 +1,18 @@
...
+```
+
+```
```
stannum-cfx, line 12: Please don't use code blocks for descriptions.
Also like Disquse said, name should be prefixed with `_`.

stannum-cfx (changes requested):

PR #512, opened 2021-07-14
Function limitation

martin951:
This function returns x=0, y=0 if called too many times within a loop. 
I've tested getting a 3D position via WORLD3D_TO_SCREEN2D and it worked.
Using set_draw_origin instead, it returned 0,0 for indexes above 20.


Disquse:
`SET_DRAW_ORIGIN`'s limit is 32.

PR #513, opened 2021-07-14
Useful tip

martin951:
Reloading a c++ script without handling the removal of those objects broke it completely.


gottfriedleibniz:
Unfortunately, these aren't bound to a `CScriptResource` like most other things. I'll expand on your comment.

PR #514, opened 2021-07-18
Update FindSpawnPointInDirection.md

martin951:


gottfriedleibniz:
Why not just update the parameters names, e.g., `float posX, float posY, float posZ, float dirX, float dirY, float dirZ` ?

martin951:
> Why not just update the parameters names, e.g., `float posX, float posY, float posZ, float dirX, float dirY, float dirZ` ?

Feel free to do that

PR #524, opened 2021-07-30
Update AddNavmeshBlockingObject.md

Factobot:


gottfriedleibniz:
Few comments:

1. Make note that only 32 "blocking" objects may exist at a given time and require manual cleanup, i.e., RemoveNavmeshBlockingObject. [CreateTrackedPoint](https://github.com/citizenfx/natives/blob/master/GRAPHICS/CreateTrackedPoint.md) can be used as a reference.
1. For C# ABI compatibility can you define `float p8` as `cs_type(Any) float p8`.
1. Move "p8 usually set to 7" to the comments section. p8 is a bitfield (1|2|4) whose meaning can be documented some other time.
1. Heading looks to be in degrees, e.g., 57.615. Also move "Heading in Radians" to the parameter section, e.g., "The heading of the block in degrees"
1. Remove code blocks from comments section and use new lines.

PR #531, opened 2021-08-04
Update SetGameplayPedHint.md p5, p6, p7 found

ziipzaaapM16A4:
p5: duration -> Time in ms to let the camera pointed at p0.
p6: blendOutDuration -> Time in ms to point the camera at p0.
p7: blendInDuration -> Time in ms to point the camera away from p0.

(Tests where made when p4 was true)
Hints taken by SET_GAMEPLAY_COORD_HINT.
Tested in GTA.

### DIFF
--- a/CAM/SetGameplayPedHint.md
+++ b/CAM/SetGameplayPedHint.md
@@ -5,7 +5,7 @@ ns: CAM
 
 ```c
 // 0x2B486269ACD548D3 0x7C27343E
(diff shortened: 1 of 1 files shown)


gottfriedleibniz (commented):
Would it be possible to rename `p5`, `p6`, and `p7` in the Parameters section?

Also `Any` -> `int`

PR #538, opened 2021-08-14
Remove broken link

0xLuca:


AvarianKnight (commented):
Link seemingly works fine, just needs be http:// (though the docs might not even support URLs)

PR #539, opened 2021-08-17
Update GetOffsetFromEntityInWorldCoords.md

draobrehtom:
This is an example implementation of `GetOffsetFromEntityInWorldCoords` which can be used on server-side.

### DIFF
--- a/ENTITY/GetOffsetFromEntityInWorldCoords.md
+++ b/ENTITY/GetOffsetFromEntityInWorldCoords.md
@@ -22,3 +22,43 @@ z = up/down
 * **offsetZ**: 
 
 ## Return value
+
+
+## Server-side implementation
+```lua
+-- server.lua
+function GetOffsetFromEntityInWorldCoords(entity, offX, offY, offZ)
+    local rot = GetEntityRotation(entity) -- ZXY
+    local rx, ry, rz = rot.x, rot.y, rot.z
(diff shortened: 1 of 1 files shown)


gottfriedleibniz:
Using Lua5.4 you can:

```lua
local glm = require('glm')
function GetOffsetFromEntityInWorldCoords(entity, offX, offY, offZ)
	local pos = GetEntityCoords(entity)
	local rot = GetEntityRotation(entity) -- ZXY
	local rx, ry, rz = math.rad(rot.x), math.rad(rot.y), math.rad(rot.z)

	-- This function would require emulation for Lua 5.3, should be quite easy to find
	local q = glm.quatEulerAngleZXY(rz, rx, ry) 

	return pos + q * vec3(offX, offY, offZ)
end
```

draobrehtom:
> Also, I'm not entirely sure how to feel about the natives repository containing fairly long/complex examples. Snippets and other utility functions should probably exist elsewhere.

I agree, but where? When I was looking for this native on server side first of all I was searching it in natives docs.

draobrehtom:
And by making such pull request I have a hope that such server native will appear on server side.

draobrehtom:
Well, actually maybe this place will be better: https://forum.cfx.re/c/general-discussion/feature-requests/

PR #540, opened 2021-08-18
Update NetworkConcealPlayer.md

itzshahay:
I've documented the NetworkConcealPlayer properly.

> This is what R* uses to hide players in MP interiors.

Did not give enough information for me to further develop my resource, where I had to research a ton, to come to the conclusion of what I've described in the PR.

There's not really a small code that would validate my change to a documentation review.


technetium-cfx:
I'd argue the original comment is a bit more accurate, given routing buckets are the appropriate solution otherwise.

itzshahay:
How would the original comment ever make any more sense? There's no literal documentation for the method, you just state what is it, not what it does or how to be utilized.

kobel240:
How would an instance be declared?

PR #542, opened 2021-08-19
Better way to make people use FiveM docs

FreakOutModz:


FreakOutModz:
@gottfriedleibniz I understand why you are doing the PR, but it automatically removes the work done by the user for the native improvisation

AvarianKnight:
You left the link in code brackets, which according to the contribution guidelines is only meant for things that were imported from the NativeDB you can see that here [here](https://github.com/citizenfx/natives/blob/master/.github/CONTRIBUTING.md#descriptions). It's much easier to go through and fix it themselves as a lot of the time when you request any changes the contributor to do so, they will take a very long time, or they won't do it at all, as can be seen [here](https://github.com/citizenfx/natives/pull/524).

FreakOutModz:
Totally agree, I had not seen on this point. Thank you for the clarification. Also I thought I had to put the link between brackets, thank you for the info!

gottfriedleibniz:
While poor git etiquette, it is much easier to squash everything, test, then apply the necessary fixes. Many of these PRs have been ignored for months now anyway.

PR #549, opened 2021-09-01
Update SetVehicleExtra.md

ToniMorton:
P2 is an INT as well as P3 so this should be explained better.

P3 is the inverse of a BOOL in this case for some reason.


technetium-cfx:
Documentation is not a place for discussion; best to keep descriptions accurate and unopinionated.

PR #562, opened 2021-09-15
Update N_0xe861d0b05c7662b8.md

TomGrobbe:


TomGrobbe:
Updated the PR to include the real native's name, thanks to alloc8or#0666

PED/SetPedCanLosePropsOnDamage.md
```diff
@@ -0,0 +1,30 @@
...
+---
+## SET_PED_CAN_LOSE_PROPS_ON_DAMAGE
+
+```c
+// 0xE861D0B05C7662B8 0x2F9550C2
+void SET_PED_CAN_LOSE_PROPS_ON_DAMAGE(Ped ped, BOOL p1, int p2);
+```
+
+It basically makes the ped lose (or not lose) their props (like glasses or helmets/hat) when someone punches or pushes the ped.
+I'm pretty sure this is what's being used in GTA:O to keep players from knocking other player's hats/glasses off when in combat.
```
blattersturm, line 13: > I'm pretty sure ..

Please refrain from writing in personal style - 'This is probably what's being used' may work fine.

PED/SetPedCanLosePropsOnDamage.md
```diff
@@ -0,0 +1,30 @@
...
+aliases: ["0xE861D0B05C7662B8"]
+---
+## SET_PED_CAN_LOSE_PROPS_ON_DAMAGE
+
+```c
+// 0xE861D0B05C7662B8 0x2F9550C2
+void SET_PED_CAN_LOSE_PROPS_ON_DAMAGE(Ped ped, BOOL p1, int p2);
+```
+
+It basically makes the ped lose (or not lose) their props (like glasses or helmets/hat) when someone punches or pushes the ped.
```
blattersturm, line 12: 'It basically' is basically filler.

PED/SetPedCanLosePropsOnDamage.md
```diff
@@ -0,0 +1,30 @@
...
+I'm pretty sure this is what's being used in GTA:O to keep players from knocking other player's hats/glasses off when in combat.
+
+This is a toggle so no need to loop it every frame.
```
blattersturm, line 15: `SET_` prefix implies this, `so no need to loop it` -> `so it doesn't need to be looped`?

blattersturm:
Third argument can be `0` or `1` (in 323, at least), and makes the native set a different flag depending on which is used.

blattersturm:
It seems the flag set by `0` prevents more props from falling off than `1` does, also matching one of the type arguments used in KNOCK_OFF_PED_PROP.

blattersturm:
Maybe `1` being used allows shooting off hats, still?

PR #563, opened 2021-09-15
Update ResetPedMovementClipset

rxyyy:
Updated the name of the second parameter of this native, and added a comment to the example provided. Also removed a reported bug caused from setting the parameter to `1.0f`, as I could not reproduce the problem with the multiple clipsets and values I tried.

### DIFF
--- a/PED/ResetPedMovementClipset.md
+++ b/PED/ResetPedMovementClipset.md
@@ -5,17 +5,15 @@ ns: PED
 
 ```c
 // 0xAA74EC0CB0AAEA2C 0xB83CEE93
-void RESET_PED_MOVEMENT_CLIPSET(Ped ped, float p1);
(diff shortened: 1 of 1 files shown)


gottfriedleibniz:
Updating `p2` to this native has been PR'd multiple times recently. All with this *horrible* parameter description:

`**transitionSpeed**: time in seconds`

So I wondered, [who](https://github.com/citizenfx/natives/commit/54f2a12b8979a30034dc669027fbac300c9c92c6#diff-96f48aa49fabf3fa823f9b64007ff2929733d6f4f9a3ee783f6d4e4b13306e9e) failed their elementary school physics class. 

... I have an excuse: that data was imported (I think. Well, I hope anyway).

rxyyy:
I just thought keeping it consistent with the other native's description of it would've been better off, I can change it if needed.

PR #570, opened 2021-09-24
Update N_0x5e657ef1099edd65.md

freedy69:
Added this native's actual name and it's purpose


MichaelCoding25:
You should probably also update the name of the actual file

freedy69:
> You should probably also update the name of the actual file

do i have to make a new pr or is there an option to edit it? i'm a bit new to this

PR #576, opened 2021-10-02
Add a new flagId

mrgharabaghi:
This flag disable random animations while the ped is in a vehicle.


gottfriedleibniz:
Presuming I'm not off-by-one... (likely)

This config bit seems to also be set `0x820E9892A77E97CD` and used in `CScenarioConditionCanPlayInCarIdle`. Coinciding with the hash collision: `0x69d28f3e = CPED_CONFIG_FLAG_CanPlayInCarIdles` .

So maybe: `0x820E9892A77E97CD = _SET_PED_CAN_PLAY_IN_CAR_IDLE` ?

PR #584, opened 2021-10-21
Update and rename N_0x632b2940c67f4ea9.md to GetScaleformMovieCursorS…

freedy69:
Documented purpose and name


HUD/GetScaleformMovieCursorSelection.md
```diff
@@ -0,0 +1,34 @@
+---
+ns: HUD
+aliases: ["0x632B2940C67F4EA9"]
+---
+## _GET_SCALEFORM_MOVIE_CURSOR_SELECTION
+ 
+```c
+// 0x632B2940C67F4EA9
+BOOL _GET_SCALEFORM_MOVIE_CURSOR_SELECTION(int scaleformHandle, bool* received, int* selectionType, int* slotIndex, int* p4);
```
stannum-cfx, line 9: When changing parameter types, please add `cs_type(Any*)` before the new type to remain backwards compatible when using C#.

stannum-cfx (changes requested):

HUD/GetScaleformMovieCursorSelection.md
freedy69: should be alright in the new commit

PR #586, opened 2021-11-03
Update and rename N_0xf06ebb91a81e09e3.md to PauseMenuSetWarnOnTabCha…

freedy69:
Added purpose and it's real name (thanks to gottfriedleibniz guessing it)


HUD/PauseMenuSetWarnOnTabChange.md
```diff
@@ -0,0 +1,15 @@
+---
+ns: HUD
```
stannum-cfx, line 2: Please add the alias for the old hash.

stannum-cfx (changes requested):

HUD/PauseMenuSetWarnOnTabChange.md
freedy69: done

> Please add the alias for the old hash.

PR #588, opened 2021-11-10
Update GetVehiclePlateType.md

Kanzia-WOT:
Added returned values for reference of others

### DIFF
--- a/VEHICLE/GetVehiclePlateType.md
+++ b/VEHICLE/GetVehiclePlateType.md
@@ -8,15 +8,19 @@ ns: VEHICLE
 int GET_VEHICLE_PLATE_TYPE(Vehicle vehicle);
 ```
 
+Returns the plates a vehicle has.
+
 ```
-Returns the plates a vehicle has:
- 0 = normal (two plates, front and back),
- 1 = unknown,
- 2 = single (only one plate, front or back),
- 3 = none (no visible plates),
- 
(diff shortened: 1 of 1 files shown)


PsychoShock (approved):
By looking into decompiled files, I have only found this returning "1" so I don't think it can return this. I have done multiples tests with vanilla cars and none has return "1". All the rest is okay :)

image

PsychoShock:
Note also that the native returns "0" when the player is not in a vehicle

gottfriedleibniz:
Reference/use the enum stored in gamecode:

```c
enum eVehiclePlateType
{
	VPT_FRONT_AND_BACK_PLATES = 0,
	VPT_FRONT_PLATES = 1,
	VPT_BACK_PLATES = 2,
	VPT_NONE = 3,
};
```

Just ensure to wrap the above in codeblocks in the markdown.

VEHICLE/GetVehiclePlateType.md
```diff
@@ -8,6 +8,15 @@ ns: VEHICLE
 int GET_VEHICLE_PLATE_TYPE(Vehicle vehicle);
 ```
 
+```
```
stannum-cfx, line 11: Please remove the code block from the description. ([example template](https://github.com/citizenfx/natives/blob/master/.github/CONTRIBUTING.md#example-documentation))

stannum-cfx (changes requested):

PR #602, opened 2021-12-31
renamed the native and documented it's purpose

freedy69:


HUD/DisableMultiplayerChat.md
```diff
@@ -0,0 +1,16 @@
+---
+ns: HUD
+aliases: ["SET_TEXT_CHAT_UNK"]
```
stannum-cfx, line 3: Alias should be prefixed with `_`

stannum-cfx (changes requested):

HUD/DisableMultiplayerChat.md
freedy69: fixed it in the new commit

PR #611, opened 2022-01-29
Add documentation for TASK_RAPPEL_DOWN_WALL

skyrossm:
I was not able to see the task progress over the network, however could be an issue with networking ropes instead of this native.

### DIFF
--- a/TASK/TaskRappelDownWall.md
+++ b/TASK/TaskRappelDownWall.md
@@ -1,26 +1,41 @@
 ---
 ns: TASK
 ---
-## _TASK_RAPPEL_DOWN_WALL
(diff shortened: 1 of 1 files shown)


TASK/TaskRappelDownWall.md
```diff
@@ -1,26 +1,41 @@
 ---
 ns: TASK
 ---
-## _TASK_RAPPEL_DOWN_WALL
+## TASK_RAPPEL_DOWN_WALL
```
stannum-cfx, line 4: This native doesn't have a Joaat hash and can therefor not be named without a `_` prefix.

stannum-cfx (changes requested):

PR #614, opened 2022-01-30
Name N_0xbc0753c9ca14b506 to HasPlayerBeenShotByACop

TheGamerzs:


PLAYER/HasPlayerBeenShotByCop.md
```diff
@@ -5,12 +5,12 @@ ns: PLAYER
 
 ```c
 // 0xBC0753C9CA14B506 0x9DF75B2A
-BOOL _0xBC0753C9CA14B506(Player player, int p1, BOOL p2);
+BOOL _HAS_PLAYER_BEEN_SHOT_BY_COP(Player player, int ms, BOOL p2);
 ```
 
 ```
```
stannum-cfx, line 11: Please format these as actual examples as described [here](https://github.com/citizenfx/natives/blob/master/.github/CONTRIBUTING.md). If you just want to document the behavior then remove this sample code from the description and remove the code block surrounding the description.

stannum-cfx (changes requested):
Please also add an alias for the old native hash for backwards compatibility.

stannum-cfx:
I'm not sure why the build is still failing, it looks good to me.

PR #616, opened 2022-02-02
Update TaskWanderInArea.md

niekschoemaker:
Biggest change would be the change of minimalLength to int.
While testing I noticed that the ped doesn't do anything at all when using a float for minimalLength, changing this to an int fixed this issue entirely.

I tested the example given, that makes the created ped wander around the area as expected.

### DIFF
--- a/TASK/TaskWanderInArea.md
+++ b/TASK/TaskWanderInArea.md
@@ -5,16 +5,45 @@ ns: TASK
 
 ```c
 // 0xE054346CA3A0F315 0xC6981FB9
-void TASK_WANDER_IN_AREA(Ped ped, float x, float y, float z, float radius, float minimalLength, float timeBetweenWalks);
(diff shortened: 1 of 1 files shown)


TASK/TaskWanderInArea.md
```diff
@@ -5,16 +5,45 @@ ns: TASK
 
 ```c
 // 0xE054346CA3A0F315 0xC6981FB9
-void TASK_WANDER_IN_AREA(Ped ped, float x, float y, float z, float radius, float minimalLength, float timeBetweenWalks);
```
stannum-cfx, line 8: This will break backwards compatibility for existing C# scripts, please annotate this type change with `cs_type(float)`

stannum-cfx (changes requested):

niekschoemaker:
Yes added the annotation, wasn't aware that was a thin

PR #617, opened 2022-02-03
Rename N_0x5f2013f8bc24ee69 to SetPlayerShortSwitchStyle

TheGamerzs:


stannum-cfx (changes requested):
Please add the original hash as an alias for backwards compatibility, see [this](https://github.com/citizenfx/natives/blob/master/.github/CONTRIBUTING.md) for an example.

PR #635, opened 2022-02-14
Update ClearPedTasksImmediately.md

TomStilson:
Added more description fo this native, and formatted using the guideline.

### DIFF
--- a/TASK/ClearPedTasksImmediately.md
+++ b/TASK/ClearPedTasksImmediately.md
@@ -8,10 +8,8 @@ ns: TASK
 void CLEAR_PED_TASKS_IMMEDIATELY(Ped ped);
 ```
 
-```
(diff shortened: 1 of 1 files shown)


TASK/ClearPedTasksImmediately.md
```diff
@@ -8,10 +8,8 @@ ns: TASK
 void CLEAR_PED_TASKS_IMMEDIATELY(Ped ped);
 ```
 
-```
-Immediately stops the pedestrian from whatever it's doing. They stop fighting, animations, etc. they forget what they were doing.  
-```
+Immediately stops the pedestrian from whatever it's doing. The difference between this and **ClearPedTasks** is that this one teleports the ped but does not change the position of the ped.
```
jaymo1011, line 11: When referring to another native, it's nice to link to it directly. To link to another native, use it's hash as the anchor target like this: `[CLEAR_PED_TASKS](#_0xE1EF3C1216AFF2CD)`

PR #639, opened 2022-02-15
Update SetWeaponDamageModifierThisFrame.md

orel56000:
This native does not need to be called every frame, IDK why someone changed its name to be this...


AvarianKnight:
This doesn't change the file name or add any aliases?

orel56000:
> This doesn't change the file name or add any aliases?

No, as I don't know how to do it, so I guess someone should fix it. even more, this should just return to what it used to be, as the name I purposed was the original, which was more correct than this.

WEAPON/SetWeaponDamageModifierThisFrame.md
```diff
@@ -19,9 +19,6 @@ Changes the weapon damage output by the given multiplier value. Must be run ever
 
 ```lua
```
stannum-cfx, line 21: Creating threads shouldn't be necessary if you're not planning on running loops that require delays.

stannum-cfx (changes requested):
File name needs to be changed accordingly, aliases also need to be fixed.
For more information, see [this](https://github.com/citizenfx/natives/blob/master/.github/CONTRIBUTING.md)

PR #640, opened 2022-02-16
Add SetVehicleOnGroundProperly missing parameter

TheGreenkey:
Add a missing parameter and the values used in scripts. Also added the return value to the return value section.

### DIFF
--- a/VEHICLE/SetVehicleOnGroundProperly.md
+++ b/VEHICLE/SetVehicleOnGroundProperly.md
@@ -5,7 +5,7 @@ ns: VEHICLE
 
 ```c
 // 0x49733E92263139D1 0xE14FDBA6
-BOOL SET_VEHICLE_ON_GROUND_PROPERLY(Vehicle vehicle);
+BOOL SET_VEHICLE_ON_GROUND_PROPERLY(Vehicle vehicle, float p1);
 ```
 
 ```
@@ -17,6 +17,8 @@ NativeDB Added Parameter 2: float p1
 ```
 
 ## Parameters
-* **vehicle**: 
+* **vehicle**: The vehicle to set on the ground properly
+* **p1**: Might correlate to the vehicle size that needs to be set on the ground. Values used in scripts: 0f, 1f, 5f, 7f, 7.5f(for bikes apparently), 8f
 
 ## Return value
+Returns `true` if the operation was successful, `false` otherwise.


jaymo1011:
Currently, as far as I am aware, changing the call signature of a native (adding new parameters) is likely to break other parts of the platform :(

stannum-cfx (changes requested):
As mentioned by Jaymo, adding new parameters is currently not supported due to backwards compatibility issues for C# resources.

TheGreenkey:
So should I revert the call signature and only mention the parameter under the parameters sections?

stannum-cfx:
> So should I revert the call signature and only mention the parameter under the parameters sections?

Correct, you can mention in the description that parameters were added/removed, but you won't be able to actually add/remove any. This would break C# backwards compatibility for scripts that are already using this native.

VEHICLE/SetVehicleOnGroundProperly.md
```diff
@@ -17,6 +17,8 @@ NativeDB Added Parameter 2: float p1
 ```
```
4mmonium, line 17: Change this to (just copy the markdown):

`````

Sets a vehicle on the ground on all wheels.

### Newly added parameters:

```
NativeDB Added Parameter 2: float p1
```

* **p1**:  Might correlate to the vehicle size that needs to be set on the ground. Values used in scripts: 0f, 1f, 5f, 7f, 7.5f(for bikes apparently), 8f.

`````

Removed the return bit since it was somewhat redundant noticing how it's already mentioned in the 'return value' section.

File: VEHICLE/SetVehicleOnGroundProperly.md [hunk 2]
```diff
@@ -17,6 +17,8 @@ NativeDB Added Parameter 2: float p1
...
 
 ## Parameters
-* **vehicle**: 
+* **vehicle**: The vehicle to set on the ground properly
+* **p1**: Might correlate to the vehicle size that needs to be set on the ground. Values used in scripts: 0f, 1f, 5f, 7f, 7.5f(for bikes apparently), 8f
```
4mmonium, line 21: Remove this parameter (p1) from the parameters section so there's no mismatch when verifying parameter count with argument declaration count declared at the top of the native.

VEHICLE/SetVehicleOnGroundProperly.md
```diff
@@ -5,7 +5,7 @@ ns: VEHICLE
 
 ```c
 // 0x49733E92263139D1 0xE14FDBA6
-BOOL SET_VEHICLE_ON_GROUND_PROPERLY(Vehicle vehicle);
+BOOL SET_VEHICLE_ON_GROUND_PROPERLY(Vehicle vehicle, float p1);
```
4mmonium, line 8: Remove `, float p1`. Basically revert it to what it was before.

4mmonium (changes requested):
Hey there, it's been a while. Requested some changes.
Thank you for your contribution! 🙂

thorium-cfx:
Closing this as it became stale

PR #641, opened 2022-02-16
drastically improve gamertag documentation

jaymo1011:
Mostly based on my own research with some help from @MegaGTAVMaster and @TomStilson.

This PR has an alternate implementation of #638.


jaymo1011:
It's surprising the checks failed the second time even though nothing changed when I re-committed to fix commit signing. Previous checks CI output: https://github.com/citizenfx/natives/runs/5223940631

stannum-cfx:
I believe the failed checks are due to the fact that you force pushed. It probably has trouble finding the correct commits. Documentation looks good to me, so I'm just going to merge it.

PR #642, opened 2022-02-17
Update IsPedSprinting.md

TomStilson:
More precise description.

### DIFF
--- a/TASK/IsPedSprinting.md
+++ b/TASK/IsPedSprinting.md
@@ -14,14 +14,15 @@ Is the ped given sprinting.
 
 ## Return value
 Whether or not the ped is sprinting. Note that if the ped is sprinting, it will return `1` instead of `true`.
-For example:
+
+Code like this:
 
 ```lua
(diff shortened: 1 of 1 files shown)


TASK/IsPedSprinting.md
```diff
@@ -7,9 +7,26 @@ ns: TASK
...
+Is the ped given sprinting.
 
 ## Parameters
-* **ped**: 
+* **ped**: Ped id.
 
 ## Return value
+Whether or not the ped is sprinting. Note that if the ped is sprinting, it will return `1` instead of `true`.
+
+Code like this:
```
stannum-cfx, line 18: Please format these as correct examples, look [here](https://github.com/citizenfx/natives/blob/master/.github/CONTRIBUTING.md) for more information. If you'd rather add more description text then you should add that above the Parameters section. This text below return value would be either added to the Return value description or ignored completely.

stannum-cfx (changes requested):

PR #656, opened 2022-03-08
Update example for use case

Rio358:


stannum-cfx (changes requested):
Please format these examples as showcased here:
https://github.com/citizenfx/natives/blob/master/.github/CONTRIBUTING.md

PR #666, opened 2022-03-29
Addition for RequestCutscene()

PappaNiels:
I was researching this native because i was trying to find the enter animation/cutscene for appartments etc. I saw there was not much documented about this native (especially the cutscene names).  So i decided to do some digging.

I found an already existing list of names by Jeremy. There are still some missing, which i am working on to add them.

The last thing I did, was adding two examples.

### DIFF
--- a/VEHICLE/GetVehicleClass.md
+++ b/VEHICLE/GetVehicleClass.md
@@ -33,6 +33,7 @@ Vehicle Classes:
 19: Military  
 20: Commercial  
 21: Trains  
+22: Open Wheel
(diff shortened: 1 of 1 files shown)


PichotM:
A [list](https://gist.github.com/PichotM/2177f5bf7ed4fd731adb6ea96902e0c1)of all cutscenes (b2545)

stannum-cfx (changes requested):
Merge conflicts should be solved as well as the description needs to be taken out of the code block.

PR #676, opened 2022-04-20
Document camera spline node smoothing

Lucas7yoshi:
Documented camera smoothing styles by pulling rendered cam coordinates throughout interpolation and graphing

### DIFF
--- a/CAM/AddCamSplineNodeUsingCamera.md
+++ b/CAM/AddCamSplineNodeUsingCamera.md
@@ -6,13 +6,19 @@ aliases: ["0x0FB82563989CF4FB"]
 
 ```c
 // 0x0FB82563989CF4FB 0xBA6C085B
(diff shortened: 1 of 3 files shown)


Lucas7yoshi:
> This will require verification, but if my partial-documentation is correct there are twenty-something 'smoothing styles' (easing functions), with only the first four being used in the script natives. See, `0x1402AA9D4` -> `0x1402689B0` (1604).
> 
> 1/2/3 look like easeInSine/easeOutSine/easeInOutSine (seems corroborated by your plots). While 23/24/25 look like easeInCirc/easeOutCirc/easeInOutCirc.

I can try plotting over 3 and see how it looks when I get home. I was intended to but got side tracked. 

I'll find a way to dump them more conveniently by just looping through every single one

Lucas7yoshi:
documented everything
should be good now

Lucas7yoshi:
resolved conflict

stannum-cfx:
Great documentation/data, thanks.

PR #680, opened 2022-04-25
Update PlayPain.md

4mmonium:
Describes what some arguments are and provides a full list of Pain IDs with their corresponding names gotten from the exe merely by hooking a game method to print the strings while feeding Pain IDs to the native (PlayPain).

### DIFF
--- a/AUDIO/PlayPain.md
+++ b/AUDIO/PlayPain.md
@@ -61,6 +61,10 @@ Below is a list of all the pain IDs (Asterisks indicate that sounds can play in 
 - 32: MELEE_LARGE_GRUNT (Nothing can be heard)
 - 33: PAIN_*_GENERIC (Low, Medium, High) 
 
+## Parameters
+* **ped**: The ped entity.
+* **painID**: The pain id, max allowed id is 33.
+
 ## Examples
 ```lua
 -- Play ON_FIRE (8) for all nearby peds, don't create an event (last parameter set to false)
@@ -68,9 +72,3 @@ for _, ped in ipairs(GetGamePool('CPed')) do
     PlayPain(ped, 8, 0.0, false)
 end
 ```
-
-## Parameters
-* **ped**: The ped entity.
-* **painID**: The pain id, max allowed id is 33.
-
-
--- a/AUDIO/PlayPain.md
+++ b/AUDIO/PlayPain.md
@@ -5,14 +5,22 @@ ns: AUDIO
 
 ```c
 // 0xBC9AE166038A5CEC 0x874BD6CB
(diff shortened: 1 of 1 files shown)


AUDIO/PlayPain.md
```diff
@@ -5,27 +5,52 @@ ns: AUDIO
 
 ```c
 // 0xBC9AE166038A5CEC 0x874BD6CB
-void PLAY_PAIN(Ped ped, int painID, int p1);
+void PLAY_PAIN(Ped ped, int painID, float p3, bool createNetEvent);
```
stannum-cfx, line 8: Adding/removing parameters is unfortunately not supported to keep C# backwards compatibility.

AUDIO/PlayPain.md
```diff
@@ -5,27 +5,52 @@ ns: AUDIO
...
-Last 2 parameters always seem to be 0.  
-EX: Function.Call(Hash.PLAY_PAIN, TestPed, 6, 0, 0);  
-Known Pain IDs  
-________________________  
-1 - Doesn't seem to do anything. Does NOT crash the game like previously said. (Latest patch)  
-6 - Scream (Short)  
-7 - Scared Scream (Kinda Long)  
-8 - On Fire  
-```
+## Description
```
stannum-cfx, line 11: The description shouldn't start with a `## Description` tag, anything in between the top most codeblock and the `## Parameters` section will be considered the description by default.

AUDIO/PlayPain.md
```diff
@@ -5,27 +5,52 @@ ns: AUDIO
...
+Plays a pain sound. A maximum of 33 pain IDs are allowed.  
+Below is a list of all the pain IDs (Asterisks indicate that sounds can play in a Low, Medium or High fashion), for example: `PAIN_LOW_GENERIC`. The corresponding pain strings belong to the game exe.
 
-```
-NativeDB Added Parameter 4: Any p3
-```
+## Pain IDs:
```
stannum-cfx, line 15: Please do not use custom `##` headers anywhere, they probably won't end up working correctly for codegen.

stannum-cfx (changes requested):

AUDIO/PlayPain.md
4mmonium: Oh ok, left the same as they were

AUDIO/PlayPain.md
4mmonium: Got rid of the description tag, for readability purposes (when people go through the native docs) I switched over to bold text style titles.

AUDIO/PlayPain.md
4mmonium: Yup, gone. Let me know if this is ok later, so I can squash the PR, thanks!

PR #696, opened 2022-04-27
Fix mix-up of IsObjectAPickup & IsObjectAPortablePickup

chipjpg:
- Looked at the checks of each native and it looks like we had the hashes mixed up
- I also compared the x360 native registration ordering with a generated PC header file and that gave me the same conclusion


blattersturm:
In both cases, this is a little fishy. `0xFC481C641EBBD27D` didn't exist until build 1365, which is quite a bit after the latest X360 TU, so this shouldn't have a 32-bit unseeded hash in the first place.

However, that indeed does not change the fact that `0x0378C08504160D0D` is indeed checking if the object is a portable pickup, as (any offsets are b323):

1. `0x0378C08504160D0D` checks for `(CObject + 1128) & 1`.
2. This is set at the end of a routine that creates a pickup (called from `CREATE_PORTABLE_PICKUP` as well as a netobj creator), which checks for `(CObject + 1104) + 296`.
3. `CObject + 1104` is initialized to the `CPickupData` instance.
4. Struct metadata (as e.g. [alexguirre's readable dump](https://raw.githubusercontent.com/alexguirre/gtav-DumpStructs/master/dumps/gtav/b2612_with_offsets.txt)) defines `0x128` (296) as `AttachmentBone`.
    ```c
    struct CPickupData {
        enum eAnimBoneTag AttachmentBone; // offset:0x128
    ```

blattersturm:
In fact, checking cs.github.com shows no notable matches for `IsObjectAPortablePickup` (in 'our' spelling), but multiple for the (wrong) IsObjectAPickup being expected to _actually_ handle any pickups.

Safe to merge, indeed.

PR #708, opened 2022-04-27
Added joaat to SeethroughSetColorNear

chipjpg:
Joaat taken from x360 TU27: https://gyazo.com/11fc2f1c626ecfc9d6bbd13add0efc77.png


GRAPHICS/SeethroughSetColorNear.md
```diff
@@ -1,12 +1,12 @@
 ---
 ns: GRAPHICS
-aliases: ["0x1086127B3A63505E"]
+aliases: ["0x1086127B3A63505E", "_SEETHROUGH_SET_COLOR_NEAR"]
```
stannum-cfx, line 3: Alias here shouldn't be necessary due to it being the same name without a `_` prefix.

stannum-cfx (changes requested):

PR #726, opened 2022-04-27
Add joaat for NetworkSetInviteOnCallForInviteMenu

chipjpg:
joaat fomr X360 TU27


NETWORK/NetworkSetInviteOnCallForInviteMenu.md
```diff
@@ -1,15 +1,15 @@
 ---
 ns: NETWORK
-aliases: ["0x66F010A4B031A331"]
+aliases: ["0x66F010A4B031A331", "_NETWORK_SET_INVITE_ON_CALL_FOR_INVITE_MENU"]
```
stannum-cfx, line 3: Also here, I don't think this alias is necessary considering it's the same name but without a `_`.

stannum-cfx (changes requested):

PR #727, opened 2022-04-27
Update NetworkCheckDataManagerSucceededForHandle.md

chipjpg:


NETWORK/NetworkCheckDataManagerSucceededForHandle.md
```diff
@@ -1,17 +1,18 @@
 ---
 ns: NETWORK
-aliases: ["0x44B37CDCAE765AAE"]
+aliases: ["0x44B37CDCAE765AAE", "_NETWORK_CHECK_DATA_MANAGER_SUCCEEDED_FOR_HANDLE"]
```
stannum-cfx, line 3: Don't think the alias here is necessary because it'd be the same name in all runtimes regardless. None of them use the prefixed `_` variant.

stannum-cfx (changes requested):

PR #732, opened 2022-04-28
Fix joaat mixup for RemoveCurrentRise

chipjpg:
Changed joaat from 0x7DBCEF6F to 0x45268B6F.
0x7DBCEF6F was used to reset the rise counter to 0 on console. afaik it's not available on new builds.


fingaweg:
0x45268B6F = REMOVE_EXTRA_CALMING_QUAD

stannum-cfx:
I'm not sure how to verify if this won't break anything, @blattersturm any ideas?

AvarianKnight:
I don't think jooat values are used for anything besides documenting if a name is valid or not, which this also fixes.

PR #736, opened 2022-04-28
Confirmed GetEntityCanBeDamaged joaat

chipjpg:
Joaat taken from x360 TU27


ENTITY/GetEntityCanBeDamaged.md
```diff
@@ -1,12 +1,12 @@
 ---
 ns: ENTITY
-aliases: ["0xD95CC5D2AB15A09F"]
+aliases: ["0xD95CC5D2AB15A09F", "_GET_ENTITY_CAN_BE_DAMAGED"]
```
stannum-cfx, line 3: I don't think adding the alias in this case is necessary because in all runtimes it'd be named the same (`GetEntityCanBeDamaged`, without the prefixing `_`).

stannum-cfx (changes requested):
pending alias removal.

PR #760, opened 2022-05-14
Import 2612 natives.

4mmonium:
Imported from https://raw.githubusercontent.com/alloc8or/gta5-nativedb-data/master/natives.json
Special thanks to alloc8or and contributors.


freedy69:
i think you might have to remove the JOAAT_UNKNOWN-s

4mmonium:
> i think you might have to remove the JOAAT_UNKNOWN-s

Yeah, I will most likely re-import without the JOAAT_UNKNOWN and just do a comment on the native somewhere instead to help others find natives with no JOAAT hashes when searching the repo :-)

PR #761, opened 2022-05-14
Update IsPedInjured description

LaFouine25:
Update description about IsPedInjured native who was false. I have add Exemple and Return Value.


PED/IsPedInjured.md
```diff
@@ -9,11 +9,19 @@ BOOL IS_PED_INJURED(Ped ped);
...
-Gets a value indicating whether this ped's health is below its injured threshold.  
-The default threshold is 100.  
+Get 1 or false if PED is Injured.
 ```
 
 ## Parameters
 * **ped**: 
 
 ## Return value
+1 or false
```
4mmonium, line 19: This function returns a boolean, therefor the return value should either be `true` or `false`. Not trying to be picky, just trying to keep things consistent :-)

PED/IsPedInjured.md
```diff
@@ -9,11 +9,19 @@ BOOL IS_PED_INJURED(Ped ped);
 ```
 
 ```
-Gets a value indicating whether this ped's health is below its injured threshold.  
-The default threshold is 100.  
+Get 1 or false if PED is Injured.
```
4mmonium, line 12: Could you remove the content out of the code block and remove the code block? This should be the description (so it shows as 'Has Modern Docs' in the native reference), code blocks are used for NativeDB imports.

You can read the contribution guide [here](https://github.com/citizenfx/natives/blob/master/.github/CONTRIBUTING.md).

I'm also not sure why you removed `Gets a value indicating whether this ped's health is below its injured threshold.`, since after all, while it can be irrelevant information for some users, it still applies, there's a function inside the native that actually checks that and that's how the game evaluates whether a ped is injured or not.  

Somewhat reversed code from 2545's exe:
```c
bool __fastcall isPedBelowInjuredThreshold(struct_pedEntity *pedEntity)
{
  _DWORD *pedHealth; // rdi
  __int64 v3; // rbx
  char v4; // al

  pedHealth = &pedEntity->health;
  v3 = pedEntity->qword290;
  v4 = 0;
  while ( v3 != pedHealth )
  {
    if ( *v3 != *pedHealth )
    {
      if ( !v4 )
      {
        nullsub();
        v4 = 1;
      }
      *v3 = *pedHealth;
    }
    v3 = *(v3 + 16);
  }
  return *pedHealth < pedEntity->injuredThreshold;
}
```

I would simply just remove everything from the code block and for the description I would just write 'Can be used to to determine whether a ped is injured or not, the game will determine this based on the ped's injured threshold.`

thorium-cfx:
Closing because there's no reply to 4mmonium's requests

PR #764, opened 2022-05-21
fix(SetPedComponentVariation): wrong link to component/props IDs

BerkieBb:
This link currently sends you to `https://docs.fivem.net/natives/gtaxscripting.blogspot.com/2016/04/gta-v-peds-component-and-props.html` on the native reference documentation, when it should send you to `https://gtaxscripting.blogspot.com/2016/04/gta-v-peds-component-and-props.html`, this PR fixes that.


BerkieBb:
I'll update the PR tomorrow

PR #766, opened 2022-05-24
Replace blogspot link for ped component and prop ids with gamecode enumeration

BerkieBb:
This replaces the link of the component and prop IDs of SetPedComponentVariation with the gamecode enum @gottfriedleibniz provided me with in #764.

### DIFF
--- a/PED/SetPedPropIndex.md
+++ b/PED/SetPedPropIndex.md
@@ -8,24 +8,40 @@ ns: PED
 void SET_PED_PROP_INDEX(Ped ped, int componentId, int drawableId, int textureId, BOOL attach);
 ```
 
-This native is used to set prop variation on a ped. Components, drawables and textures IDs are related to the ped model. 
+This native is used to set prop variation on a ped. Components, drawables and textures IDs are related to the ped model.
 
 ### MP Freemode list of props
-**0**: Hat  
-**1**: Glass  
-**2**: Ear  
-**6**: Watch  
-**7**: Bracelet  
+**0**: Hats
+**1**: Glasses
+**2**: Ears
+**6**: Watches
+**7**: Bracelets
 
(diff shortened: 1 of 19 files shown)


AvarianKnight:
Might also be a good idea to link IsPedComponentVariationValid to this instead of blogspot

BerkieBb:
> Might also be a good idea to link IsPedComponentVariationValid to this instead of blogspot

Done

AvarianKnight (approved):

BerkieBb:
That should be all blogspot links removed

PED/SetPedPropIndex.md
```diff
@@ -8,24 +8,40 @@ ns: PED
...
+**7**: Bracelets
 
-### Related and useful natives
-[GET_NUMBER_OF_PED_PROP_DRAWABLE_VARIATIONS](#_0x5FAF9754E789FB47)  
-[GET_NUMBER_OF_PED_PROP_TEXTURE_VARIATIONS](#_0xA6E7F1CEB523E171)  
-
-[List of component/props ID](https://gtaxscripting.blogspot.com/2016/04/gta-v-peds-component-and-props.html) of player_two with examples
+List of Prop IDs
+```c
+// Props
```
4mmonium, line 22: I was very confused at why this was here when this is not needed by the native at first, then I realized that this data is referenced by many other related natives. Perhaps this would be better off in some section within the FiveM docs with a reference to it, but I guess it's fine here for now.

4mmonium:
I think this PR is good to go 😃

PED/SetPedPropIndex.md
BerkieBb: Yeah probably in the `Game references` section of the docs would be the best

PR #767, opened 2022-05-27
Include language that informs native should be used per frame

jnccloud:
This native, unlike some others that are appended with "ThisFrame," does not have that luxury. This should help reduce forum topics complaining that the native does not work.

### DIFF
--- a/PLAYER/DisablePlayerVehicleRewards.md
+++ b/PLAYER/DisablePlayerVehicleRewards.md
@@ -7,10 +7,9 @@ ns: PLAYER
 // 0xC142BE3BB9CE125F 0x8C6E611D
 void DISABLE_PLAYER_VEHICLE_REWARDS(Player player);
 ```
-```
+
 Disables vehicle rewards for the current frame.  
-```
 
 ## Parameters
-* **player**: 
+* **player**: The player to disable rewards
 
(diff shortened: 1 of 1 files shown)


PLAYER/DisablePlayerVehicleRewards.md
```diff
@@ -7,7 +7,9 @@ ns: PLAYER
 // 0xC142BE3BB9CE125F 0x8C6E611D
 void DISABLE_PLAYER_VEHICLE_REWARDS(Player player);
 ```
-
+```
```
AvarianKnight, line 10: Code blocks are meant for NativeDB imports

AvarianKnight (changes requested):

PLAYER/DisablePlayerVehicleRewards.md
jnccloud: Was in the process of removing them. Is this better? (My first time contributing to the docs, sorry!)

AvarianKnight (approved):

4mmonium:
Was looking for a citation stating this, turns out there's this:
https://forum.cfx.re/t/disableplayervehiclerewards-does-not-disable-vehicle-rewards-properly/1764536/4

LGTM :-)

nikez:
While we're at it; possible to add an icon/tag on natives to indicate ThisFrame-requirements? It's really a nice to have and not a need to have kind of feature.

PR #769, opened 2022-06-05
Update SetPlayerResetFlagPreferRearSeats.md

niekschoemaker:
This native is only used in the mobile phone scripts of GTA, and in these scripts there's a switch case on a global which contains phone settings.

This same global field is also passed to a scaleform method called "SET_THEME", and to a function which changes the object variation of the mobile phone object, as such I have concluded that this is probably something related to the theme settings of the mobile phone.

In game this native also appears to have no effect whatsoever, which would line up with it being related to the mobile phone.

### DIFF
--- a/PLAYER/SetPlayerResetFlagPreferRearSeats.md
+++ b/PLAYER/SetPlayerResetFlagPreferRearSeats.md
@@ -12,7 +12,8 @@ void SET_PLAYER_RESET_FLAG_PREFER_REAR_SEATS(Player player, int flags);
 example:  
 flags: 0-6  
 PLAYER::SET_PLAYER_RESET_FLAG_PREFER_REAR_SEATS(PLAYER::PLAYER_ID(), 6);  
-wouldnt the flag be the seatIndex?  
+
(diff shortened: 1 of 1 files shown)


niekschoemaker:
I also understand that this doesn't really clear anything up functionality wise, but just want to avoid confusion for other people, as the native name can be quite confusing.

blattersturm:
Hash collision, perhaps -> the name ought to be removed then?

4mmonium:
A quick search on the offset led me to find that undocumented native [`0x83A169EABCDB10A2`](https://github.com/citizenfx/natives/blob/fdb8fccd026770491f80abd19f48d84cbb7f9fb6/PED/N_0x83a169eabcdb10a2.md?plain=1#L4) follows the same pattern, but for peds, only difference is how it accesses the ped by using `getNetPlayerPed` instead of `getPedFromScriptEntity`.

PsychoShock:
This native is just wrong param's. Seem's to be the ped index and the vehicle index. R* set sometimes vehicles index to NULL to apply to all vehicles.

From what I can see, it enables to gently encourage players to prioritize entering the rear seats of the designated vehicle or every vehicles if second param set for this.

PR #778, opened 2022-06-20
SetEntityCompletelyDisableCollision -> SetEntityCompletelyEnableCollision

orel56000:
If you set the toggle argument to true, it collides, if set to false, it has no collisions, which means the current name is the opposite of what it does.

This PR does not affect the actual code; If someone can do it, it will be blessed!


4mmonium (changes requested):
Maybe instead of renaming the native, give it a description as you did in the commit message, and explain every parameter (fill them), that way the users who read this native doc will know it's actual behavior regardless of naming, thanks 😀

4mmonium:
> You never change natives that are confirmed, instead put it in the description that it does something different.

I read your comment right after posting mine, sorry 😅

PR #781, opened 2022-06-22
Update GetScriptTaskStatus.md

MadsLeander:
I've added some information regarding what the native returns as well as an example in lua.

I've also made a small video showing what the example code shows: https://www.youtube.com/watch?v=jqTgp0d0tOI


MadsLeander:
I've done some more research and it seems you are right about the  `SCRIPT_TASK_ANY`, it seems to return 1 when doing a task and 3 when not doing any task. `SCRIPT_TASK_INVALID` works similarly, but returns 7 if there are any ongoing tasks, and 3 if there are no ongoing tasks.

I suppose `VACANT_STAGE = 3` does fit with the results described above. 
`DORMANT_TASK = 2` also seems to be a more fitting label. (instead of "is getting hindered")

I'll be doing some more testing to see if I can draw some more parallels and possibly prove your theory.

MadsLeander:
I've updated the PR, it includes some documentation on `SCRIPT_TASK_ANY` and `SCRIPT_TASK_INVALID`, as well as a list of the confirmed returns and their labels. This is where I'll end it for now, unless someone else has something more to add.

PR #807, opened 2022-09-01
Document more camera natives.

4mmonium:
Document CreateCamWithParams (add explanations for active, rotationOrder and completely
re-write camName parameter example), DoesCamExist, GetCamRot and
GetGameplayCamRot.

Correct a typo with GetCamRot, document SetCamRot

### DIFF
--- a/CAM/DoesCamExist.md
+++ b/CAM/DoesCamExist.md
@@ -8,11 +8,10 @@ ns: CAM
 BOOL DOES_CAM_EXIST(Cam cam);
 ```
 
-```
-Returns whether or not the passed camera handle exists.  
-```
+Looks up a camera handle in the current camera pool and returns `true` if the handle is found, otherwise it returns `false`.
 
 ## Parameters
-* **cam**: 
+* **cam**: The camera handle to look for.
 
 ## Return value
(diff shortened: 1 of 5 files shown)


CAM/DoesCamExist.md
```diff
@@ -8,11 +8,10 @@ ns: CAM
 BOOL DOES_CAM_EXIST(Cam cam);
 ```
 
-```
-Returns whether or not the passed camera handle exists.  
-```
+Looks up a camera handle in the current camera pool and returns `True` if the handle is found, otherwise it returns `False`.
```
AvarianKnight, line 11: Extreme nit pick, but is there a reason to capitalize True and False?

4mmonium: > Extreme nit pick, but is there a reason to capitalize True and False?

Huh, that's a good point. I have always seen it be capitalized, I guess this is to blame?
https://learn.microsoft.com/en-us/dotnet/api/system.boolean.tostring?view=net-7.0

I can make it lowercase.

4mmonium: Done & squashed! 🙂

AvarianKnight (approved):

PR #813, opened 2022-09-19
IsPlayerDrivingDangerously

Lucas7yoshi:
0xF10B44FD479D69F3 is used for a voice line about the player driving in "the correct lanes" (wrong lanes) for nigel2.c


themonthofjune:
GET_IS_PLAYER_DRIVING_WRECKLESS

Lucas7yoshi:
> 

I will not apply this due to the origin of it unless I get told to do otherwise

4mmonium (changes requested):
I looked up the native in IDA and `p2` may be some sort of type, it ranges from `0` to `2`.

Apparently `2` is against traffic as you're saying, but then I looked up other offsets that the native checks depending on `p2` and there's a check for traffic lights. I noticed it was traffic lights since I did some research on junctions and traffic lights a while ago and that offset led me there. Pseudo-code provided for demonstration purposes only.

Notice the comment (names are speculative guesses):
image

And  `p2` being `1` matches `0xbf8`, which is the check for traffic lights, game code specifically checks if a traffic light is red.
image

So we know two things for sure:
`State/Type 1`: Is a check for traffic lights, specifically red.
`State/Type 2`: Is a check for driving against traffic as you are saying.
`State/Type 0`: Is a check for 'last time driven on pavement', may get set when the player isn't driving on pavement/car is in the air?

Lucas7yoshi:
Testing reveals that 1 does indeed seem to be for running through red lights, does take a bit of time for it to begin to return true though (which might have something to do with the time mentions there?)

Testing with value 0 i could not find any situation in which it would return true.

4mmonium:
> Testing reveals that 1 does indeed seem to be for running through red lights, does take a bit of time for it to begin to return true though (which might have something to do with the time mentions there?)
> 
> Testing with value 0 i could not find any situation in which it would return true.

0 is the time since the player drove on pavement, check this out:
image

Notice `0x1078` is the exact same offset in the other screenshot I posted [here](https://github.com/citizenfx/natives/pull/813#pullrequestreview-1730699158)

Then there's the `0xBF4` offset which confirms it:
image

Maybe not driving on pavement or having the car fly in the air causes the native to return something when calling the `IsPlayerDrivingDangerously` native with `type` `0`?

Lucas7yoshi:
> > Testing reveals that 1 does indeed seem to be for running through red lights, does take a bit of time for it to begin to return true though (which might have something to do with the time mentions there?)
> > Testing with value 0 i could not find any situation in which it would return true.
> 
> 0 is the time since the player drove on pavement, let's check this native and follow the line: image
> 
> Notice `0x1078` is the exact same offset in the other screenshot I posted [here](https://github.com/citizenfx/natives/pull/813#pullrequestreview-1730699158)
> 
> Then there's the `0xBF4` offset which confirms it: image
> 
> Maybe not driving on pavement or having the car fly in the air causes the native to return something when calling the `IsPlayerDrivingDangerously` native with `type` `0`?

i believe i tried that while testing all 3 and got nothing. might be non functional

JajarGaming:
> > 
> 
> I will not apply this due to the origin of it unless I get told to do otherwise

https://nativedb.dotindustries.dev/gta5/natives/0xF10B44FD479D69F3?search=0xF10B44FD479D69F3+

PR #818, opened 2022-09-20
Update N_0x3e38e28a1d80ddf6.md to IsControlledVehicleUnableToGetToRoad.md

QuadrupleTurbo:


titanium-cfx:

PR #819, opened 2022-09-27
Update ClearArea* of Cops, Peds and Projectiles.

4mmonium:
This PR modifies an incorrectly named parameter in the aforementioned natives. 
The parameter was previously named `flags`, but this parameter actually defines whether a net event of `CClearAreaEvent` should be created or not upon native execution.

I'm not entirely sure if changing parameter types is ok to do so (as in won't break anything, never done such thing), if so I can just set the type back to integer, let me know if so, thanks.

### DIFF
--- a/MISC/ClearAreaOfCops.md
+++ b/MISC/ClearAreaOfCops.md
@@ -5,17 +5,21 @@ ns: MISC
 
 ```c
 // 0x04F8FC8FCF58F88D 0x95C53824
-void CLEAR_AREA_OF_COPS(float x, float y, float z, float radius, int flags);
+void CLEAR_AREA_OF_COPS(float x, float y, float z, float radius, cs_type(int) BOOL createNetEvent);
 ```
 
-```
-flags appears to always be 0
+Clears an area of cops at the given coordinates and radius.
+
+## Examples
+```lua
+-- Clear the area, do also create an event
+ClearAreaOfCops(0.0, 0.0, 0.0, 10000.0, true)
 ```
 
 ## Parameters
-* **x**: 
-* **y**: 
-* **z**: 
-* **radius**: 
-* **flags**: 
+* **x**: The x coordinate of where to clear cops.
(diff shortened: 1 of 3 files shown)


4mmonium:
> Don't arguments with their type changed need a cs_type to maintain compatibility with CS?

Ah right, my bad.
https://github.com/citizenfx/natives/pull/819/files#diff-763fc62b945d7d5993004da239fe99184962d59f4ff0a6b181b4c248f16336eaL8

4mmonium:
Ok, I believe I corrected it. I didn't do it for [this](https://github.com/citizenfx/natives/pull/819/files#diff-763fc62b945d7d5993004da239fe99184962d59f4ff0a6b181b4c248f16336eaL8) native since it was previously done before, so I just switched it to the right type.

Anyhow, [this](https://github.com/citizenfx/fivem/blob/44c6199b68829df79313bfb20aeb09ce68fafd37/ext/natives/codegen_out_cs.lua#L211) is where it parses the parameters.

MISC/ClearAreaOfCops.md
```diff
@@ -5,17 +5,22 @@ ns: MISC
...
 ```c
 // 0x04F8FC8FCF58F88D 0x95C53824
-void CLEAR_AREA_OF_COPS(float x, float y, float z, float radius, int flags);
+void CLEAR_AREA_OF_COPS(float x, float y, float z, float radius, cs_type(int) BOOL createNetEvent);
 ```
 
-```
-flags appears to always be 0
+Clears an area of cops at the given coordinates and radius.
+`createNetEvent` specifies whether a `CClearAreaEvent` event of should be created for online use.
```
AvarianKnight, line 12: Is this line necessary considering its already defined in the parameters? (same goes for the other natives)

4mmonium: I just corrected it, thanks!

AvarianKnight (approved):
Lgtm

PR #801, opened 2022-08-25
Correct some issues caused by an older PR

4mmonium:
Corrects some issues present in the following files:

- CreateCam.md
- AddBlipForEntity.md
- AddBlipForRadius.md
- NetworkGetNetworkIdFromEntity.md
- CreateObject.md

This should also address some issues with non-working js examples.

PR #821, opened 2022-10-03
Update SetSuperJumpThisFrame and other natives

4mmonium:
This PR documents:

- SetSuperJumpThisFrame
- GetPlayersLastVehicle
- GetVehicleCauseOfDestruction

A user on Discord asked for source code regarding
`SetSuperJumpThisFrame` I believe they didn't know how to use the native, so I documented it anyway.

Upon documenting `GetVehicleCauseOfDestruction` (since the given example was some code from the game's decompiled scripts) I noticed that `GetPlayersLastVehicle` was of no use if the vehicle to get was previously destroyed, so I documented on that too.

### DIFF
--- a/MISC/SetSuperJumpThisFrame.md
+++ b/MISC/SetSuperJumpThisFrame.md
@@ -8,7 +8,24 @@ ns: MISC
 cs_type(int) void SET_SUPER_JUMP_THIS_FRAME(Player player);
 ```
 
+Allows the player to perform super jumps. This function must be called every frame for it to work.
+It basically OR's a flag for a single frame, allowing the ped to perform a super jump only when the flag is set.
+
+
 ## Parameters
-* **player**: 
+* **player**: The player we are setting this for.
 
 ## Return value
+
+## Examples
+```lua
+Citizen.CreateThread(function()
+    while true do
+        SetSuperJumpThisFrame(PlayerId())
+        -- Try enabling the two down below if you also want invincibility and no ragdoll effects
+        --SetPlayerInvincible(PlayerId(), true)
+        --SetPedCanRagdoll(PlayerPedId(), false)
+        Citizen.Wait(0)
+    end
+end)
+```
--- a/PLAYER/GetPlayersLastVehicle.md
+++ b/PLAYER/GetPlayersLastVehicle.md
@@ -8,8 +8,11 @@ ns: PLAYER
 Vehicle GET_PLAYERS_LAST_VEHICLE();
 ```
 
-```
-Alternative: GET_VEHICLE_PED_IS_IN(PLAYER_PED_ID(), 1);  
-```
+### Warning
+This native will return `0` if the last vehicle the player was in was destroyed.
+
+### Alternative
+You can use [GET_VEHICLE_PED_IS_IN](#_0x9A9112A0FE9A4713), which will actually get the last vehicle, even if it was destroyed.
 
 ## Return value
+A vehicle handle containing the last player's vehicle.
--- a/VEHICLE/GetVehicleCauseOfDestruction.md
+++ b/VEHICLE/GetVehicleCauseOfDestruction.md
@@ -8,16 +8,21 @@ ns: VEHICLE
 Hash GET_VEHICLE_CAUSE_OF_DESTRUCTION(Vehicle vehicle);
 ```
 
-```
-iVar3 = get_vehicle_cause_of_destruction(uLocal_248[iVar2]);  
-if (iVar3 == joaat("weapon_stickybomb"))  
-{  
-     func_171(726);  
-     iLocal_260 = 1;  
-}  
-```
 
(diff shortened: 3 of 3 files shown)


MISC/SetSuperJumpThisFrame.md
```diff
@@ -8,7 +8,47 @@ ns: MISC
...
 ```
 
+Allows the player to perform super jumps. This function must be called every frame for it to work.
+It basically OR's a flag for a single frame, allowing the ped to perform a super jump only when the flag is set.
+
+This flag does also get set in other native functions, for instance 
+
+[SET_EXPLOSIVE_AMMO_THIS_FRAME](#_0xA66C71C98D5F2CFB), [SET_FIRE_AMMO_THIS_FRAME](#_0x11879CDD803D30F4) and [SET_EXPLOSIVE_MELEE_THIS_FRAME](#_0xFF1BED81BFDC0FE0) do also set this flag.
+
+You can check if a player is using **Super Jump** by calling [IS_PLAYER_USING_SUPER_JUMP](#_0xC7D2C20C) from the server.
```
AvarianKnight, line 18: Not sure if this is really useful information to put onto a client native

MISC/SetSuperJumpThisFrame.md
```diff
@@ -8,7 +8,47 @@ ns: MISC
...
+  ped = (struct_PedData *)getPlayerPedFromHandle(Player);
+  if ( ped )
+  {
+    // Offsets are commented, pedSpecialFlags gets accessed and modified down below
+    ped = (struct_PedData *)ped->pedInfo;       // *(QWORD*)(ped + 0x1078)
+    ped->pedSpecialFlags |= 0x4000u;                   // *(DWORD*)(ped + 0x150)
+  }
+  return ped;
+}
+```
```
AvarianKnight, line 36: This is probably useless information to put in here as most people won't know what it even means

PLAYER/GetPlayersLastVehicle.md
```diff
@@ -8,8 +8,11 @@ ns: PLAYER
 Vehicle GET_PLAYERS_LAST_VEHICLE();
 ```
 
-```
-Alternative: GET_VEHICLE_PED_IS_IN(PLAYER_PED_ID(), 1);  
-```
+Will return the player's last vehicle. The native will return `0` if the last vehicle the player was in was destroyed.
+
+This native accesses a global variable which gets unset by the game when a vehicle is destroyed (also gets unset in many other places), so it's not useful for cases where vehicles are destroyed.
```
AvarianKnight, line 13: Redundant information

VEHICLE/GetVehicleCauseOfDestruction.md
```diff
@@ -8,16 +8,23 @@ ns: VEHICLE
...
-```
-iVar3 = get_vehicle_cause_of_destruction(uLocal_248[iVar2]);  
-if (iVar3 == joaat("weapon_stickybomb"))  
-{  
-     func_171(726);  
-     iLocal_260 = 1;  
-}  
-```
+Allows you to get a vehicle's cause of destruction. 
+Down below (in the examples section) we can find an example that determines if our vehicle was destroyed by a sticky bomb or not.
```
AvarianKnight, line 12: Redundant information, the native name describes what it does

AvarianKnight (commented):
Most of this is just nit picky stuff

MISC/SetSuperJumpThisFrame.md
4mmonium: Yeah, this was just to address the 0.00001 percent that asked for 'source code', but yeah, most people won't even know the underlying meaning of it, will remove.

PLAYER/GetPlayersLastVehicle.md
4mmonium: Will remove the redundancy, but will still warn that users shouldn't use this native if they plan to access a 'last vehicle' that was destroyed by who knows 😀

4mmonium:
> Most of this is just nit picky stuff

Thanks, removed all the extra unneeded stuff end users may not want to see.

PLAYER/GetPlayersLastVehicle.md
```diff
@@ -8,8 +8,11 @@ ns: PLAYER
 Vehicle GET_PLAYERS_LAST_VEHICLE();
 ```
 
-```
-Alternative: GET_VEHICLE_PED_IS_IN(PLAYER_PED_ID(), 1);  
-```
+### Warning
```
AvarianKnight, line 11: Are you able to add additional sections like this? I thought it would break native gen

4mmonium: Yes 😀
https://github.com/citizenfx/natives/pull/801#discussion_r955428610

titanium-cfx:
LGTM

PR #824, opened 2022-10-16
fix(vehicle/IsVehicleWindowIntact): eWindowId enum

BerkieBb:
I have tested the natives IS_VEHICLE_WINDOW_INTACT, ROLL_WINDOW_DOWN, ROLL_WINDOW_UP, SMASH_VEHICLE_WINDOW and FIX_VEHICLE_WINDOW and after going through the indexes 0-7 I have found that the order of the enum on the docs weren't correct, so this PR fixes that


4mmonium (approved):
LGTM, tested this using `RemoveVehicleWindow`.

PR #826, opened 2022-11-02
Update SetEnableHandcuffs.md

MegaMagnum:

### DIFF
--- a/PED/SetEnableHandcuffs.md
+++ b/PED/SetEnableHandcuffs.md
@@ -9,7 +9,9 @@ void SET_ENABLE_HANDCUFFS(Ped ped, BOOL toggle);
 ```
 
 ```
-ped can not pull out a weapon when true  
+Ped cannot pull out a weapon
+Ped will instantly die when in contact with water
+When ped stands up from ragdoll it will have cuffed hands
 ```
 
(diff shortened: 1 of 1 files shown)


PED/SetEnableHandcuffs.md
```diff
@@ -9,7 +9,9 @@ void SET_ENABLE_HANDCUFFS(Ped ped, BOOL toggle);
 ```
 
 ```
-ped can not pull out a weapon when true  
+Ped cannot pull out a weapon
```
4mmonium, line 12: Markdown [code blocks](https://www.markdownguide.org/extended-syntax/#fenced-code-blocks) are reserved for nativedb imports. If you're adding documentation to a native, get rid of the code blocks altogether, here's a [link](https://github.com/citizenfx/natives/blob/master/.github/CONTRIBUTING.md) to the contribution guide :-)

The reason for this is that it will mark the [native documentation](https://docs.fivem.net/natives/) as 'Has "imported" docs', when you actually contributed to it, when it should be 'Has "modern" docs' (when the description isn't left blank).

Thanks!

4mmonium (changes requested):

PR #830, opened 2022-11-09
Add missing argument and improve documentation

niekschoemaker:
I'm not 100% sure as to what the arguments do exactly because the behavior of the native changes depending on whether oal is enabled or not, not sure what's going on there, I'm assuming this is due to the missing argument.

Without OAL it always does the non flashing fade in, and with OAL it always does the flashing fade in, even though this is not the actual missing argument.

### DIFF
--- a/NETWORK/NetworkFadeInEntity.md
+++ b/NETWORK/NetworkFadeInEntity.md
@@ -5,18 +5,19 @@ ns: NETWORK
 
 ```c
 // 0x1F4ED342ACEFE62D 0x9B9FCD02
-void NETWORK_FADE_IN_ENTITY(Entity entity, BOOL flash, BOOL slow);
+void NETWORK_FADE_IN_ENTITY(Entity entity, BOOL flash);
 ```
 
 Fade the given entity back in, usually used after the entity has been faded out with [NETWORK_FADE_OUT_ENTITY](#_0xDE564951F95E09ED)
 
 When used on a entity which isn't invisible or faded out then the native will still work, it will just instanly make the ped invisible before fading in.
 
+Note that there is another boolean parameter which is missing, when set to true the fade in will be a little bit slower.
+
 
 ## Parameters
 * **entity**: The entity to fade in
 * **flash**: When set to true the entity will flash rapidly while fading out
-* **slow**: When set to true the fadein will be a little bit slower than normal
 
 ## Examples
 
@@ -28,6 +29,7 @@ end
 
 --- Do something like a teleport, or warp into a vehicle
 
+-- Last parameter here is slow, this parameter is missing in decleration
 NetworkFadeInEntity(PlayerPedId(), false, false)
 while NetworkIsEntityFading(PlayerPedId()) do
   Citizen.Wait(0)
--- a/NETWORK/NetworkFadeInEntity.md
+++ b/NETWORK/NetworkFadeInEntity.md
@@ -5,14 +5,15 @@ ns: NETWORK
 
 ```c
 // 0x1F4ED342ACEFE62D 0x9B9FCD02
-void NETWORK_FADE_IN_ENTITY(Entity entity, BOOL flash);
+void NETWORK_FADE_IN_ENTITY(Entity entity, BOOL bNetwork);
 ```
 
 Fade the given entity back in, usually used after the entity has been faded out with [NETWORK_FADE_OUT_ENTITY](#_0xDE564951F95E09ED)
 
 When used on a entity which isn't invisible or faded out then the native will still work, it will just instanly make the ped invisible before fading in.
 
-Note that there is another boolean parameter which is missing, when set to true the fade in will be a little bit slower.
+**Additional Parameters**:
+* **flash**: If set to true the entity will flash while fading in.
 
 
 ## Parameters
--- a/NETWORK/NetworkFadeInEntity.md
+++ b/NETWORK/NetworkFadeInEntity.md
@@ -5,7 +5,7 @@ ns: NETWORK
 
 ```c
 // 0x1F4ED342ACEFE62D 0x9B9FCD02
-void NETWORK_FADE_IN_ENTITY(Entity entity, BOOL state);
+void NETWORK_FADE_IN_ENTITY(Entity entity, BOOL flash, BOOL slow);
 ```
 
 Fade the given entity back in, usually used after the entity has been faded out with [NETWORK_FADE_OUT_ENTITY](#_0xDE564951F95E09ED)
--- a/NETWORK/NetworkFadeInEntity.md
+++ b/NETWORK/NetworkFadeInEntity.md
@@ -18,21 +18,24 @@ When used on a entity which isn't invisible or faded out then the native will st
 
 ## Parameters
 * **entity**: The entity to fade in
-* **flash**: When set to true the entity will flash rapidly while fading out
+* **network**: When set to true the fade in will be networked.
 
 ## Examples
 
 ```lua
-NetworkFadeOutEntity(PlayerPedId(), false, false)
-while NetworkIsEntityFading(PlayerPedId()) do
-  Citizen.Wait(0)
+local playerPed = PlayerPedId()
+
+-- fade out the player while flashing them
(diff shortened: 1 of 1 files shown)


niekschoemaker:
I've validated that the argument is indeed missing from decompiled scripts, all calls I could find where in the form of:
`NETWORK::NETWORK_FADE_IN_ENTITY(iVar0, true, 1);`

NETWORK/NetworkFadeInEntity.md
```diff
@@ -5,20 +5,31 @@ ns: NETWORK
 
 ```c
 // 0x1F4ED342ACEFE62D 0x9B9FCD02
-void NETWORK_FADE_IN_ENTITY(Entity entity, BOOL state);
+void NETWORK_FADE_IN_ENTITY(Entity entity, BOOL flash, BOOL slow);
```
4mmonium, line 8: At this time adding new arguments in this fashion in not supported. Maybe make a note of it in the native description? Thanks!

NETWORK/NetworkFadeInEntity.md
```diff
@@ -5,20 +5,31 @@ ns: NETWORK
...
-```
-NativeDB Added Parameter 3: BOOL slow
-```
 
 ## Parameters
-* **entity**: 
-* **state**: 
+* **entity**: The entity to fade in
+* **flash**: When set to true the entity will flash rapidly while fading out
+* **slow**: When set to true the fadein will be a little bit slower than normal
```
4mmonium, line 19: Remove once you remove the native function declaration argument mentioned up above or it won't compile.

4mmonium (changes requested):
Looks good, except for the added argument you declared, that's currently not supported. Requested some changes, thank you for your contribution 😊

niekschoemaker:
> Looks good, except for the added argument you declared, that's currently not supported. Requested some changes, thank you for your contribution 😊

I've ammended the requested changes to the pull request, wasn't entirely sure which part could be done, but removed any "official" reference to the parameter and added a comment in the example and in the description about the missing parameter.

NETWORK/NetworkFadeInEntity.md
```diff
@@ -5,20 +5,33 @@ ns: NETWORK
...
-```
-state - 0 does 5 fades  
-state - 1 does 6 fades  
-native is missing third argument, also boolean, setting to 1 made vehicle fade in slower, probably "slow" as per NETWORK_FADE_OUT_ENTITY  
-```
+Fade the given entity back in, usually used after the entity has been faded out with [NETWORK_FADE_OUT_ENTITY](#_0xDE564951F95E09ED)
+
+When used on a entity which isn't invisible or faded out then the native will still work, it will just instanly make the ped invisible before fading in.
+
+Note that there is another boolean parameter which is missing, when set to true the fade in will be a little bit slower.
```
4mmonium, line 15: I actually did a full research on this on Sunday, the last parameter is basically a `bSendOverNet` boolean param which basically specifies if the fade should be sent over the network or not.

NETWORK/NetworkFadeInEntity.md
```diff
@@ -5,20 +5,33 @@ ns: NETWORK
...
 ## Parameters
-* **entity**: 
-* **state**: 
+* **entity**: The entity to fade in
+* **flash**: When set to true the entity will flash rapidly while fading out
+
+## Examples
+
+```lua
+NetworkFadeOutEntity(PlayerPedId(), false, false)
```
4mmonium, line 25: `NetworkFadeOutEntity` has the boolean parameters reversed, first one is `bSendOverNet` and second one is `bFlash` or `flash`, I just append `b` behind it to specify it's a boolean.

I highlighted the diffs in red.
diffs

What I also noticed is that fading an entity with occupants inside, such as a vehicle, will also fade the occupants, but only if they're classified as net objects (which is basically a net object method that sets the alpha for every entity within the seat manager instance of the specific vehicle entity)

image

I improved your current example in order to demonstrate it:
```lua
RegisterCommand("fademe", function(source, args, rawCommand)
    CreateThread(function()
        local entity = GetVehiclePedIsIn(PlayerPedId())
        if entity == 0 then
            entity = PlayerPedId()
        else
            if IsVehicleSeatFree(entity, 0) then
                local playerDummy = ClonePed(PlayerPedId(), true, false, true)
                SetPedIntoVehicle(playerDummy, entity, 0)
            end
        end
       -- Last param is send over net, second is flash
        NetworkFadeOutEntity(entity, false, true)
        while NetworkIsEntityFading(entity) do
            Citizen.Wait(0)
        end
        -- Second parameter here is bSendOverNet, last one is flash
        NetworkFadeInEntity(entity, true, false)
        while NetworkIsEntityFading(entity) do
            Citizen.Wait(0)
        end
    end)
end, false)
```

Note: screenshots are provided for illustration/educational purposes only.

4mmonium (changes requested):
Requested changes, but this time we know for sure what the parameters are 😊
Apologies for not researching this in full-detail earlier.

NETWORK/NetworkFadeInEntity.md
```diff
@@ -5,20 +5,33 @@ ns: NETWORK
 
 ```c
 // 0x1F4ED342ACEFE62D 0x9B9FCD02
-void NETWORK_FADE_IN_ENTITY(Entity entity, BOOL state);
+void NETWORK_FADE_IN_ENTITY(Entity entity, BOOL flash);
```
AvarianKnight, line 8:
```suggestion
void NETWORK_FADE_IN_ENTITY(Entity entity, BOOL bNetwork);
```

NETWORK/NetworkFadeInEntity.md
```diff
@@ -5,20 +5,33 @@ ns: NETWORK
...
 
-```
-NativeDB Added Parameter 3: BOOL slow
-```
 
 ## Parameters
-* **entity**: 
-* **state**: 
+* **entity**: The entity to fade in
+* **flash**: When set to true the entity will flash rapidly while fading out
```
AvarianKnight, line 20:
```suggestion
* **network**: When set to true the entity will flash rapidly while fading out
```

NETWORK/NetworkFadeInEntity.md
AvarianKnight, line 15:
```suggestion
**Additional Parameters**:
* **flash**: If set to true the entity will flash while fading in.
```

NETWORK/NetworkFadeInEntity.md
```diff
@@ -5,20 +5,33 @@ ns: NETWORK
...
+end
 
+--- Do something like a teleport, or warp into a vehicle
+
+-- Last parameter here is slow, this parameter is missing in decleration
+NetworkFadeInEntity(PlayerPedId(), false, false)
+while NetworkIsEntityFading(PlayerPedId()) do
+  Citizen.Wait(0)
+end
+```
```
AvarianKnight, line 37:
```suggestion
local playerPed = PlayerPedId()

-- fade out the player while flashing them
NetworkFadeOutEntity(playerPed, true, false)
while NetworkIsEntityFading(playerPed) do
  Wait(0)
end

--- Do something like a teleport, or warp into a vehicle

-- while generally frowned upon when you can use natives, this declaration has
-- a missing parameter, which doesn't work without manually invoking.
Citizen.InvokeNative(0x1F4ED342ACEFE62D, playerPed, false, true)
while NetworkIsEntityFading(playerPed) do
  Wait(0)
end
```

AvarianKnight (changes requested):

NETWORK/NetworkFadeInEntity.md
```diff
@@ -5,20 +5,37 @@ ns: NETWORK
...
 
-```
-NativeDB Added Parameter 3: BOOL slow
-```
 
 ## Parameters
-* **entity**: 
-* **state**: 
+* **entity**: The entity to fade in
+* **network**: When set to true the fade in will be networked.
```
AvarianKnight, line 21:
```suggestion
* **bNetwork**: When set to true the fade in will be networked.
```

PR #831, opened 2022-11-10
Update SetVehicleExtra.md

r0adra93:
The question was bugging me because Computers operate in an on or off state. Yes or No, True or False.
I discovered that the reason this method is backwards due to the fact it is a flag that hides the extra instead of shows the extra. Hide Extra = 0 or 1

### DIFF
--- a/VEHICLE/SetVehicleExtra.md
+++ b/VEHICLE/SetVehicleExtra.md
@@ -10,14 +10,13 @@ void SET_VEHICLE_EXTRA(Vehicle vehicle, int extraId, BOOL disable);
 
 ```
 Note: only some vehicle have extras  
-extra ids are from 1 - 9 depending on the vehicle  
+extra ids are from 1 - 13 depending on the vehicle 
+-------------------------------------------------
+The method asks a question: Hide Extra?
+0 answers No 
+1 answers Yes
 -------------------------------------------------  
-^ not sure if outdated or simply wrong. Max extra ID for b944 is 14  
--------------------------------------------------  
-p2 is not a on/off toggle. mostly 0 means on and 1 means off.  
-not sure if it really should be a BOOL.  
--------------------------------------------------  
-Confirmed p2 does not work as a bool. Changed to int. [0=on, 1=off]  
+Confirmed p2 does not work as a bool. Changed to int.
 ```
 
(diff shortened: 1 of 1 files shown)


VEHICLE/SetVehicleExtra.md
```diff
@@ -10,14 +10,13 @@ void SET_VEHICLE_EXTRA(Vehicle vehicle, int extraId, BOOL disable);
 
 ```
 Note: only some vehicle have extras  
-extra ids are from 1 - 9 depending on the vehicle  
+extra ids are from 1 - 13 depending on the vehicle 
+-------------------------------------------------
+The method asks a question: Hide Extra?
+0 answers No
```
4mmonium, line 16: I don't want to sound pedantic, but it's currently declared as a Bool in the native method, so while it's true `1` and `0` represent bit on/off states, it should be `False` or `True` to conform with the native method declaration.

Also, if you're improving a native, it's better to just take the content out of the code fences and declare it as the description, since code fences are reserved for documentation coming from native imports and [native docs](https://docs.fivem.net/natives) will incorrectly mark it as "Has "imported" docs" instead of "Has "modern" docs" even though you made changes to it. 

There's more on the [contribution](https://github.com/citizenfx/natives/blob/master/.github/CONTRIBUTING.md) guide.

Thanks for your contribution though! Maybe check on those tiny details I pointed out 🙂

For reference, here's the decompiled game native from 323

```c
int __fastcall sub_7FF7032FCCE0(signed int vehicle, int extraId, char disable)
{
  __int64 v5; // rax
  __int64 v6; // rbx

  v5 = getCVehicleFromScriptHandle(vehicle);
  v6 = v5;
  if ( v5 )
  {
    if ( (extraId - 1) <= 13 )
    {
      LODWORD(v5) = sub_7FF70355358C(v5, extraId + 290, disable);
      if ( disable )
      {
        if ( !*(v6 + 2680) )
        {
          LODWORD(v5) = sub_7FF70349A18C(v6, 0i64);
          if ( !v5 )
          {
            sub_7FF7034A4930(v6, 42i64);
            LODWORD(v5) = sub_7FF7034A4930(v6, 43i64);
          }
        }
      }
    }
  }
  return v5;
}
```

PR #832, opened 2022-11-11
[FIX] hash will never be a string

daiguel:
added GetHashKey() because hash will never be a string always a number


Disquse:
Your change will make this code invalid. Please read this: https://cookbook.fivem.net/2019/06/23/lua-support-for-compile-time-jenkins-hashes/

PR #834, opened 2022-11-16
Update StopAnimTask.md

orel56000:


TASK/StopAnimTask.md
```diff
@@ -14,5 +14,5 @@ void STOP_ANIM_TASK(Ped ped, char* animDictionary, char* animationName, float p3
 * **ped**: 
 * **animDictionary**: 
 * **animationName**: 
-* **p3**:
```
4mmonium, line 17: If `p3` is `blendOutSpeed` as you are saying, then the argument should be set as such as well (see [STOP_ANIM_TASK](https://github.com/citizenfx/natives/pull/834/files#diff-d0d165f2281fd505d6880864665e17de1b371d04fcd6fba5310442ae44570da7L8) native declaration). Additionally, you can't put spaces under parameters, it causes the tests to fail.

Thanks for your contribution though!

4mmonium (changes requested):

TASK/StopAnimTask.md
orel56000: Oh, I see. Thanks!

4mmonium: Hey there, this is still wrong, you named the parameter `Blend out speed`, but it should be named `blendOutSpeed`, _make sure it matches the function parameter name, otherwise it will still fail to verify_. Thanks!

thorium-cfx:
Closing this as it became stale

PR #837, opened 2022-11-29
Update IsVehicleSirenOn and GetVehicleDoorAngleRatio

Flohhhhh:
Expected this to return false or true, instead returns false or 1.

Tested with prints.

```lua
CreateThead(function()
    while true do
        local veh = GetVehiclePedIsIn(PlayerPedId())
        print("[Lights Check] " .. tostring(IsVehicleSirenOn(veh)))
        Wait(1000)
    end
end)
```

image

### DIFF
--- a/VEHICLE/IsVehicleSirenOn.md
+++ b/VEHICLE/IsVehicleSirenOn.md
@@ -13,7 +13,7 @@ Returns whether the vehicle's lights and sirens are on.
 * **vehicle**: 
 
 ## Return value
-Returns false if siren and lights are disabled or 1 if they enabled.
+Returns false if siren and lights are disabled or true if they are enabled.
 
 ## Examples
 ```lua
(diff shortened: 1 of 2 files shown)


Flohhhhh:
First PR let me know if I screwed something : D

logan-mcgee:
i want to say from memory, that this is intended/normal behaviour for CfxLua, and unrelated to the native used, but i could be wrong

AvarianKnight:
It is for compatibility iirc, as long as you don't do specific `== true` or `== false` it shouldn't matter anyways

VEHICLE/IsVehicleSirenOn.md
```diff
@@ -7,9 +7,18 @@ ns: VEHICLE
...
 BOOL IS_VEHICLE_SIREN_ON(Vehicle vehicle);
 ```
-
+Returns whether the vehicle's lights and sirens are on.
 
 ## Parameters
 * **vehicle**: 
 
 ## Return value
+Returns false if siren and lights are disabled or 1 if they enabled.
```
Flohhhhh, line 16: Typo 😭 
"if they enabled" > "if they're enabled"

logan-mcgee: the return value shouldnt be 1 anyway, as mentioned its purely for compatability, its just a truthy value and used correctly the value doesnt matter, its just true/false

Flohhhhh:
Corrected typo and change from ``1`` to ``true``
Thanks everyone for the advice.

PR #842, opened 2022-12-12
Fix references for weather natives

niekschoemaker:
All weather natives referred to SET_WEATHER_TYPE_NOW_PERSIST which referred to SET_WEATHER_TYPE_NOW (which was a broken link to begin with)

Instead of doing that I made all natives refer directly to SET_WEATHER_TYPE_NOW.


4mmonium (approved):
Looks good to me, thanks!

PR #845, opened 2022-12-24
Update NetworkFadeInEntity.md

freedy69:
Added a new parameter and also an example

### DIFF
--- a/NETWORK/NetworkFadeInEntity.md
+++ b/NETWORK/NetworkFadeInEntity.md
@@ -5,20 +5,14 @@ ns: NETWORK
 
 ```c
 // 0x1F4ED342ACEFE62D 0x9B9FCD02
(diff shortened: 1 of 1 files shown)


NETWORK/NetworkFadeInEntity.md
```diff
@@ -5,20 +5,14 @@ ns: NETWORK
 
 ```c
 // 0x1F4ED342ACEFE62D 0x9B9FCD02
-void NETWORK_FADE_IN_ENTITY(Entity entity, BOOL state);
```
4mmonium, line 8: If I recall correctly, you can't add parameters like this. There are some compatibility constraints with the ABI, hence why there's a fenced comment down below reading `NativeDB Added Parameter 3: BOOL slow`.

4mmonium (changes requested):
Check what I commented about the added parameters, thanks!

PR #846, opened 2022-12-26
Fix typo

TGBFIX1:
Saw a typo while looking for some features to make, figured why not change it


STATS/StatSetBool.md
```diff
@@ -10,7 +10,7 @@ BOOL STAT_SET_BOOL(Hash statName, BOOL value, BOOL save);
 
 ```
 Example:
- STATS::STAT_SET_BOOL(MISC::GET_HASH_KEY("MPPLY_MELEECHLENGECOMPLETED"), trur, true);
+ STATS::STAT_SET_BOOL(MISC::GET_HASH_KEY("MPPLY_MELEECHLENGECOMPLETED"), true, true);
```
4mmonium, line 13: Even better, why not just create a lua example using `lua` code fences i.e.

```lua
StatSetBool(GetHashKey('MPPLY_MELEECHLENGECOMPLETED'), true, true)
```

4mmonium (changes requested):
Check what I commented, thank you 🙂

PR #848, opened 2023-01-07
fix(AttachEntityToEntity): Changed the last parameter to match reality

XenoS-ITA:
The last parameter means exactly the opposite, if it is false the rotation does not update on the attached entity, otherwise it updates along with the entity. Before it wrote fixedRot, which implied the opposite (if true the rotation will be fixed)


AvarianKnight:
You seem to have forgotten to fix the name in the function declaration, which is why the CI failed

XenoS-ITA:
Sorry but i have never changed native parameters, could you please guide me?

XenoS-ITA:
Sorry but i hadn't seen those parameters 😅

PR #852, opened 2023-01-12
Update a few Bink Movie Natives

ItsJunction:
Accidentally deleted the repository of my previous PR, was quite messy though so here's cleaner.

### DIFF
--- a/GRAPHICS/DrawBinkMovie.md
+++ b/GRAPHICS/DrawBinkMovie.md
@@ -12,8 +12,8 @@ Must be called each frame, will play at specified position on screen when called
 
 ## Parameters
 * **binkMovie**: The movie to be drawn (from [`_SET_BINK_MOVIE`](#_0x338D9F609FD632DB)).
-* **posX**: The centered x position of the movie. (0.0 - 1.0).
-* **posY**: The centered y position of the movie. (0.0 - 1.0).
+* **posX**: The centered x position of the movie (0.0 - 1.0).
+* **posY**: The centered y position of the movie (0.0 - 1.0).
 * **scaleX**: The x scale of the movie (0.0 - 1.0).
 * **scaleY**: The y scale of the movie (0.0 - 1.0).
 * **rotation**: The rotation of the movie (0.0 - 360.0).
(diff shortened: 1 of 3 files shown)


GRAPHICS/DrawBinkMovie.md
```diff
@@ -5,21 +5,34 @@ ns: GRAPHICS
 
 ```c
 // 0x7118E83EEB9F7238
-void _DRAW_BINK_MOVIE(int binkMovie, float p1, float p2, float p3, float p4, float p5, int r, int g, int b, int a);
+void _DRAW_BINK_MOVIE(int binkMovie, float posx, float posy, float scalex, float scaley, float rotation, int r, int g, int b, int a);
```
4mmonium, line 8: Looks like parameter names are case sensitive, in other words, if your function parameters don't match your `## Parameters` section, PR validation checks will fail, i.e.

Naming a parameter `posx` when it should be `posX` as defined in the ` ## Parameters` description section.

```
ParseError: A parameter definition was found for a parameter that does not exist. Parameter name: posX
```

4mmonium (changes requested):
There were some issues while trying to validate your PR, check what I commented, thanks! 😃

```
ParseError: A parameter definition was found for a parameter that does not exist. Parameter name: posX
```

GRAPHICS/DrawBinkMovie.md
```diff
@@ -5,21 +5,34 @@ ns: GRAPHICS
...
-* **p2**:
-* **p3**:
-* **p4**:
-* **p5**:
-* **r**:
-* **g**:
-* **b**:
-* **a**:
+* **binkMovie**: The movie that is being played.
+* **posX**: The centered x position of the movie.
```
4mmonium, line 15: Sorry I forgot to mention earlier, albeit obvious, perhaps it should be specified that this is a percentage that ranges from 0.0 to 1.0.

Just a quick nitpick really.

Otherwise this pretty much looks good to me in its current state 😃

4mmonium (approved):
Looks good to me for the most part, examples too. Thank you for your contribution.

PR #853, opened 2023-01-13
Added params for `SETUP_FAKE_CONE_DATA`

PappaNiels:
This pull request will add some of the parameters of this native (which is still a hash). Can be used for cctv cams or maybe even guards.

### DIFF
--- a/HUD/N_0xf83d0febe75e62c9.md
+++ b/HUD/N_0xf83d0febe75e62c9.md
@@ -5,20 +5,42 @@ ns: HUD
 
 ```c
 // 0xF83D0FEBE75E62C9
-void _0xF83D0FEBE75E62C9(Any p0, Any p1, Any p2, Any p3, Any p4, Any p5, Any p6, Any p7);
+void _0xF83D0FEBE75E62C9(cs_type(Any) Blip blip, Any p1, Any p2, cs_type(Any) float width, Any p4, cs_type(Any) float length, cs_type(Any) float heading, Any p7);
 ```
 
 ```
-NativeDB Added Parameter 9: Any p8
+NativeDB Added Parameter 9: int hudColour
 NativeDB Introduced: v1290
 ```
 
+Sets the 'data' for a cone for a blip
+
+See [here](https://docs.fivem.net/docs/game-references/hud-colors/) for the hud colours for the cone
+
+Name in scripts: `SETUP_FAKE_CONE_DATA()`
+
+For people who don't know how to convert degrees to radians and the other way around:
+
+```
+degrees = radians * 180 / pi
+radians = degrees * pi / 180
+```
+
 ## Parameters
-* **p0**:
-* **p1**:
-* **p2**:
-* **p3**:
-* **p4**:
-* **p5**:
-* **p6**:
-* **p7**:
+* **blip**: A blip from an entity. The entity can not be a ped. The width and length will not change if it is a ped.
+* **p1**: -1.0 in decompiled scripts (unk)
+* **p2**: 1.0 in decompiled scripts (unk)
+* **width**: Width of the cone (0.0 to ~3.0) 
+* **p4**: 1.0 in decompiled scripts (unk)
+* **length**: Length of the cone. (8.2 is 'normal' sized)
+* **heading**: This heading is in radians, this can be calculated with (heading + 180.00) * (pi/180)
+* **p7**: 1 in decompiled scripts (unk)
+
+## Examples
+
+```lua
+blip = AddBlipForEntity(--[[entity]])
+-- ...
+Citizen.InvokeNative(0xf83d0febe75e62c9, blip, -1.0, 1.0, 0.36, 1.0, 8.2, (0.5 * math.pi), 0, 11)
+SetBlipShowCone(blip, true, 11)
+```


HUD/N_0xf83d0febe75e62c9.md
```diff
@@ -5,20 +5,27 @@ ns: HUD
...
-* **p7**:
+* **blip**: A blip from an entity, for example ch_prop_ch_cctv_cam_01a
+* **p1**: -1.0 in decompiled scripts (unk)
+* **p2**: 1.0 in decompiled scripts (unk)
+* **width**: Width of the cone (0.0 to ~3.0) 
+* **p4**: 1.0 in decompiled scripts (unk)
+* **length**: Length of the cone. (8.2 is 'normal' sized
+* **heading**: Not your typical heading. Can be calculated with (heading + 180.00) * 0.017453292
+* **p7**: 1 in decompiled scripts (unk)
+* **p8**: Hud colour
```
4mmonium, line 31:
```
ParseError: A parameter definition was found for a parameter that does not exist. Parameter name: p8
```

At the time, new parameters cannot be added, if they were indeed added to the native method then they should be defined in the native description, look at the following native method for an example:

https://github.com/citizenfx/natives/blob/e7ffedc4a33afe3a68600a649a4f641442217478/VEHICLE/GetVehicleHealthPercentage.md?plain=1#L13

Thank you for your contribution nonetheless!

4mmonium (changes requested):
There are some validation issues, take a look at what I commented 🙂

HUD/N_0xf83d0febe75e62c9.md
PappaNiels: I’ll fix this as soon as I can

PappaNiels:
Did it in the correct 'style' as the native you mentioned. Should be good now :)

HUD/N_0xf83d0febe75e62c9.md
```diff
@@ -5,20 +5,26 @@ ns: HUD
...
-* **p5**:
-* **p6**:
-* **p7**:
+* **blip**: A blip from an entity, for example ch_prop_ch_cctv_cam_01a
+* **p1**: -1.0 in decompiled scripts (unk)
+* **p2**: 1.0 in decompiled scripts (unk)
+* **width**: Width of the cone (0.0 to ~3.0) 
+* **p4**: 1.0 in decompiled scripts (unk)
+* **length**: Length of the cone. (8.2 is 'normal' sized
+* **heading**: Not your typical heading. Can be calculated with (heading + 180.00) * 0.017453292
```
4mmonium, line 29: That `0.017453292` seems to be `π/180` as in Pi divided by 180, a.k.a. `3.14159265359/180`

PappaNiels: Rotation in radians, huh? This is (for me) the easiest approach and doesn’t over complicate it. If needed ill edit it _again_ and commit it

4mmonium: > Rotation in radians, huh? This is (for me) the easiest approach and doesn’t over complicate it. If needed ill edit it _again_ and commit it

Yeah, now I recall that the game itself does this for other native methods as in degree to radian conversions or so. Anyhow, this is good, thanks!

HUD/N_0xf83d0febe75e62c9.md
```diff
@@ -5,20 +5,26 @@ ns: HUD
 
 ```c
 // 0xF83D0FEBE75E62C9
-void _0xF83D0FEBE75E62C9(Any p0, Any p1, Any p2, Any p3, Any p4, Any p5, Any p6, Any p7);
+void _0xF83D0FEBE75E62C9(Blip blip, Any p1, Any p2, float width, Any p4, float length, float heading, Any p7);
```
4mmonium, line 8: I almost forgot, when changing argument types, you should add `cs_type (Any)` annotations to whatever it is you're changing, it's for backward compatibility with old C# function parameter types (those that are changed).

Take a look at this example:
https://github.com/citizenfx/natives/blob/ba66873f1167eef2f4dfdcb7e4803d88264ccae7/PATHFIND/SetGpsDisabledZone.md?plain=1#L8

So all in all you would get:

```c
void _0xF83D0FEBE75E62C9(cs_type(Any) Blip blip, Any p1, Any p2, cs_type(Any) float width, Any p4, cs_type(Any) float length, cs_type(Any) float heading, Any p7);
```

Apologies for all the constant back and forth, thank you for your contribution.

PappaNiels: This is not needed for the hash native or does it? I'll add this nonetheless.

4mmonium: > This is not needed for the hash native or does it? I'll add this nonetheless. Thanks for the solution!

Good question, only if return types are changed, but in this case the function type is still `void`, so no.
I'll approve this PR now, thanks! 😃

4mmonium (approved):
Looks good to me, thank you for your contribution!

PappaNiels:
Added the conversion for rad to deg and deg to rad and mentioned it for the param like thorium said

HUD/N_0xf83d0febe75e62c9.md
```diff
@@ -19,12 +19,19 @@ See [here](https://docs.fivem.net/docs/game-references/hud-colors/) for the hud
...
+
 ## Parameters
 * **blip**: A blip from an entity, for example ch_prop_ch_cctv_cam_01a
 * **p1**: -1.0 in decompiled scripts (unk)
 * **p2**: 1.0 in decompiled scripts (unk)
 * **width**: Width of the cone (0.0 to ~3.0) 
 * **p4**: 1.0 in decompiled scripts (unk)
 * **length**: Length of the cone. (8.2 is 'normal' sized
-* **heading**: Not your typical heading. Can be calculated with (heading + 180.00) * 0.017453292
+* **heading**: Not your typical heading. This is in radians. Can be calculated with (heading + 180.00) * (pi/180)
```
AvarianKnight, line 36: "This heading is in radians, this can be calculated with (heading + 180.00) * (pi/180)" might be better wording

PappaNiels: welp this is one of my weird habits of doing _exactly_ what is being told to me and dropping my own 'quality control' :/

HUD/N_0xf83d0febe75e62c9.md
gottfriedleibniz: For niche[1] undocumented natives like this, this is generally not required.

`Any` would be mapped to `int`, so if a server is successfully using the native definition (without manually invoking), they are jumping through some serious hoops. And in the infinitely small case of that being untrue, it should serve as punishment for not documenting their research.

[1] The rule of thumb I used in the past was: undocumented with more complex types or requirements (e.g., the value needs to be non-zero so false/0/0.0, etc., conflicting is meaningless).

...

Radians are a universal unit, including the formula for converting between degrees and radians is superfluous ('... in radians' as a parameter description should be sufficient). Secondly, [Tau](https://tauday.com/tau-manifesto) is superior.

PappaNiels:
Any update on if the latest changes have been approved and when it will be merged?

4mmonium:
> Any update on if the latest changes have been approved and when it will be merged?

This looks good to me, I don't think it requires any other changes, at least from what I have seen here. I can't really give ETA's on merges though since I don't handle that, but if most reviewers approve, it will most likely get merged 🙂

PappaNiels:
Ah ok. We will play the wait game then. Thanks for your response!

HUD/N_0xf83d0febe75e62c9.md
```diff
@@ -5,20 +5,33 @@ ns: HUD
...
 
 ```
-NativeDB Added Parameter 9: Any p8
+NativeDB Added Parameter 9: int hudColour
 NativeDB Introduced: v1290
 ```
 
+Sets the 'data' for a cone for a blip
+
+See [here](https://docs.fivem.net/docs/game-references/hud-colors/) for the hud colours for the cone
```
thorium-cfx, line 18: Mixed British and American spelling (color/colour), do we apply a standard for docs? (not method signatures)

HUD/N_0xf83d0febe75e62c9.md
```diff
@@ -5,20 +5,33 @@ ns: HUD
...
+
+Name in scripts: `SETUP_FAKE_CONE_DATA()`
+
+For people who don't know how to convert degrees in radians and the other way around:
```
thorium-cfx, line 22: convert degrees _**to**_ radians

HUD/N_0xf83d0febe75e62c9.md
```diff
@@ -5,20 +5,33 @@ ns: HUD
...
 ## Parameters
-* **p0**:
-* **p1**:
-* **p2**:
-* **p3**:
-* **p4**:
-* **p5**:
-* **p6**:
-* **p7**:
+* **blip**: A blip from an entity, for example ch_prop_ch_cctv_cam_01a
```
thorium-cfx, line 30: The example of a string (piece of text) is misguiding for type `Blip` (integer base type)

HUD/N_0xf83d0febe75e62c9.md
```diff
@@ -5,20 +5,33 @@ ns: HUD
...
+* **p1**: -1.0 in decompiled scripts (unk)
+* **p2**: 1.0 in decompiled scripts (unk)
+* **width**: Width of the cone (0.0 to ~3.0) 
+* **p4**: 1.0 in decompiled scripts (unk)
+* **length**: Length of the cone. (8.2 is 'normal' sized
```
thorium-cfx, line 35: Missing end of the parenthesis `)`

thorium-cfx (changes requested):
Some grammar/spelling and an example mistake, apart from that it looks good to me.

HUD/N_0xf83d0febe75e62c9.md
4mmonium: > Mixed British and American spelling (color/colour), do we apply a standard for docs? (not method signatures)

I'm not really sure, even native names seem to mix these😅
I prefer the AmE spelling though, searched for it and found more occurrences.

HUD/N_0xf83d0febe75e62c9.md
4mmonium: > The example of a string (piece of text) is misguiding for type `Blip` (integer base type)

Nice catch I completely missed this, I tried to find the unknown `p*` args in the game code to see if this could be complemented a bit, but no joy.

thorium-cfx: Ah I think I understand it now; we first need to get the blip id that's related to this "ch_prop_ch_cctv_cam_01a" object through some other means. If so, maybe explaining that process or leaving it out?

HUD/N_0xf83d0febe75e62c9.md
thorium-cfx: Ah yeah with method signatures I was referring to the natives, seems like R* is also not sure which one to pick, lol.
I also prefer AmE, but more so sticking to one or the other :)

HUD/N_0xf83d0febe75e62c9.md
4mmonium: > Ah I think I understand it now; we first need to get the blip id that's related to this "ch_prop_ch_cctv_cam_01a" object through some other means. If so, maybe explaining that process or leaving it out?

I think an easier way to explain it could be:
A blip from an entity, for example the player's entity or a vehicle entity, no? 😃

PappaNiels: The behavior on peds is weird. Thats why the ai blips exist for them. Rather exclude it than have a ‘native not working properly’, right?

thorium-cfx: Would something like `GetBlipFromEntity(entity)` be used? Something like that would explain it pretty well. Can also make a small note that peds have undefined behavior.

PappaNiels: Is possible. Or `AddBlipForEntity()` (not a ped). But for peds the width, etc doesnt work sadly

thorium-cfx (approved):

thorium-cfx:
This probably got lost in the sea of work. @4mmonium can you merge if everything's still good with this one?

PappaNiels:
New reminder @4mmonium :) . Would be unfortunate if this research got lost.

PsychoShock (commented):
Hello,

After some tests/research, this is what I have. 

I think the current `p1` and `p2` unknow is the minimum/maximum horizontal F.O.V. for the ped (I think any value wouldn't affect the cone blip render on the minimap). I have try with [`-1.0, 1.0` ](https://i.imgur.com/ZKk7IM0.png) and [`-10.0, 10.0`](https://i.imgur.com/VK6w1vB.png) and results seems to be the same for the cone. Perhaps @4mmonium have better tools to verify this.

Seeing the decompiled files, I found the `width` goes up to `1.0` max only according to `fm_content_club_management.c`.

I can't seem to found anything for `p4` and `p7` (wich is **always** `true/1`).

-----------------------------------------------
So if I follow this, it would be like this:

[SetFakeConeData.md](https://github.com/citizenfx/natives/files/15083327/SetFakeConeData.md)


This is just my personnal opinion. Doesn't help much, but maybe someone will found something better based on my research.

Thanks :)

4mmonium:
Hi, I'm merging this 😊

PR #857, opened 2023-01-28
Update CONTRIBUTING.md

4mmonium:
Let users know practices on renaming parameters, changing parameter types, to avoid entering values without any actual documentation behind it and to not use double hashes inside description blocks.

I mainly updated the [CONTRIBUTING.md](https://github.com/citizenfx/natives/blob/master/.github/CONTRIBUTING.md) file with this since it's something I noticed other users doing when contributing documentation.

Let me know if I missed anything, or if there's other points which could be improved, thanks!


AvarianKnight:
LGTM 👍

titanium-cfx:
LGTM

PR #858, opened 2023-01-30
switch p3 & p6

Datsun-Drift:
Upon testing with a bool variable determined by checking bone distances, p3 and p6 are labeled backwards.

targetRear is the bool I define from the dist checks AttachVehicleToTowTruck(veh,target,0.0,0.0,0.0,targetRear)

When using as published before, the connection would only go to the rear and the hook was completely away from the car

### DIFF
--- a/VEHICLE/AttachVehicleToTowTruck.md
+++ b/VEHICLE/AttachVehicleToTowTruck.md
@@ -5,7 +5,7 @@ ns: VEHICLE
 
 ```c
 // 0x29A16F8D621C4508 0x8151571A
-void ATTACH_VEHICLE_TO_TOW_TRUCK(Vehicle towTruck, Vehicle vehicle, BOOL rear, float hookOffsetX, float hookOffsetY, float hookOffsetZ);
+void ATTACH_VEHICLE_TO_TOW_TRUCK(Vehicle towTruck, Vehicle vehicle, float hookOffsetX, float hookOffsetY, float hookOffsetZ, BOOL rear);
 ```
 
 ```
@@ -14,9 +14,9 @@ HookOffset defines where the hook is attached. leave at 0 for default attachment
 
 ## Parameters
 * **towTruck**: 
-* **vehicle**: 
-* **rear**: 
+* **vehicle**:  
 * **hookOffsetX**: 
 * **hookOffsetY**: 
 * **hookOffsetZ**: 
+* **rear**:
 


AvarianKnight:
This would break backwards compatibility with old scripts, I'm also not sure if this is even correct, in GTA V decompiled scripts its never called with a boolean last, or even a boolean at all (its always -1)

```c
VEHICLE::ATTACH_VEHICLE_TO_TOW_TRUCK(towTruck, targetVehicle, -1, 0f, 0f, 0f);
```

4mmonium (changes requested):
I reviewed your changes and decided to disassemble the native method, it seems that your changes are also wrong and they may work due to pure coincidence.

The original parameter order seems to be correct, but the parameter `rear` declared as a `BOOL` type seems to be wrong, it should be a `boneIndex` of type `int` and you will see why when we look at the game's native pseudo-code:

```c
void __fastcall VEHICLE::ATTACH_VEHICLE_TO_TOW_TRUCK_ACTUAL(native args)
{
  scrNativeCallContextArgList *hookOffsetVecXYZ; // rdi
  int boneIndex; // er14
  Entity vehicleEnt; // ebx
  CEntity *towTruckEntity; // rsi
  CEntity *vehicle; // rax
  int hookOffsetY; // xmm1_4
  int hookOffsetZ; // xmm2_4
  int vehiclePtr; // ebp
  signed int attachPointIdx; // ebx
  CVehicleTrailerAttachPoint *vehTrailerAttachPoint; // rdi
  scrNativeCallContextArgStructP *nArgs; // rax
  _DWORD hookOffsetXYZ[10]; // [rsp-28h] [rbp-28h] BYREF

  nArgs = args->pArgs;
  hookOffsetVecXYZ = &nArgs->p3;
  boneIndex = nArgs->p2.Any;
  vehicleEnt = nArgs->p1.Any;
  towTruckEntity = getEntityAddressIfVehicle(nArgs->p0.Any);
  vehicle = getEntityAddressIfVehicle(vehicleEnt);
  hookOffsetY = hookOffsetVecXYZ[1].Any;
  hookOffsetZ = hookOffsetVecXYZ[2].Any;
  vehiclePtr = vehicle;
  hookOffsetXYZ[0] = hookOffsetVecXYZ->Any;
  hookOffsetXYZ[1] = hookOffsetY;
  hookOffsetXYZ[2] = hookOffsetZ;
  if ( towTruckEntity )
  {
    if ( vehicle )
    {
      attachPointIdx = 0;
      if ( LOWORD(towTruckEntity[13].ModelInfo) )
      {
        do
        {
          vehTrailerAttachPoint = getVehicleTrailerAttachPoint(towTruckEntity, attachPointIdx);
          if ( (vehTrailerAttachPoint->vtable_0->SYSTEM::_LOG_DEBUG_INFO_IMPL_copy_7)(vehTrailerAttachPoint) == 18 )
            sub_140EE455C(vehTrailerAttachPoint, towTruckEntity, vehiclePtr, boneIndex, hookOffsetXYZ, 0);
          ++attachPointIdx;
        }
        while ( attachPointIdx < LOWORD(towTruckEntity[13].ModelInfo) );
      }
    }
  }
}
```

Now, if we look at a specific part of the following game method `sub_140EE455C` (this is from 2545), we can see that there's a check being done to see if the `boneId` we passed to it is not `-1`, so this pretty much confirms it's not a `BOOL` type at all.

```c
if ( boneIndex != -1 )
  {
    v48 = *(a3 + 32);
    v261 = v48;
    if ( v48 )
    {
      if ( CalledFromNative_2249(v48) && *(CalledFromNative_2249(v261) + 24) )
      {
        v49 = CalledFromNative2_2683(v261);
        v47 = au_re_boneIdAsBoneIndex_1(v49, 0, boneIndex);
LABEL_30:
        if ( v47 == -1 )
          v47 = 0;
        goto LABEL_32;
      }
    }
  }
```

Now one can get the `boneIndex` by calling [GetEntityBoneIndexByName](https://docs.fivem.net/natives/?_0xFB71170B7E76ACBA).

So it indeed looks like this has been wrong for a long while... I indeed confirmed it was a `boneIndex` by looking at the game's disassembly for [GetPedBoneIndex](https://docs.fivem.net/natives/?_0x3F428D08BE5AAE31) and noticing it called the exact game method this is calling declared as:

```c
boneIdAsBoneIndex(crSkeletonData *pBoner, BoneId16 boneId, BoneIndex *pBoneIndex)
```

So all in all `BOOL rear` may have to be changed to an integer type named `boneIndex`.

I also tested this using `InvokeNative`:

```c
Citizen.InvokeNative( 0x29A16F8D621C4508, towTruck, vehicleToAttach, GetEntityBoneIndexByName(vehicleToAttach, "chassis"), 0.0, 0.0, 0.0 )
```

PR #861, opened 2023-02-05
Update CreatePickup.md description

Glowable:
The original description of CreatePickup.md has an invalid link that needs to be updated that used to show the names of Pickup Hashes.
I have created a new pastebin with the new Pickup Hashes, which actually works.


blattersturm:
Replacing a link that died with another link that might very well die the same way - not too useful. 

Adding a list to the docs (citizenfx/fivem-docs) instead, and referencing that one, would make more sense.

Glowable:
I'll add the links to the docs soon.

PR #862, opened 2023-02-06
Enable compat generation

thorium-cfx:

Executes the compat generator for C# native compatibility methods.

Notes:
1. requires thorium-cfx/native-doc-tooling@14e7b346542a1690335730be1dbd04288d7f2c42 before it can be activated.
2. this was tested with reverting commit citizenfx/native-doc-tooling@1968b2ea01ff1006806519a5bb39246150f78b0b, as this broke the native generator and was also apparently not checked out in the natives repo, this needs some attention first.
3. initial run for the compatibility file should've been been run and its output file needs to be present on the remote server. This can be done manually or by making a separate GitHub Action.
4. rsync's `NATIVE_DEPLOY_URL` constants may need a good check as I don't know the format of its contents.


blattersturm:
> initial run for the compatibility file should've been been run and its output file needs to be present on the remote server. This can be done manually or by making a separate GitHub Action.

How should this be done, exactly?

thorium-cfx:
> How should this be done, exactly?

The simplest would be to execute `node compatgen.js -i --start-date=2023-01-01` inside the natives repo directory. GitHub's fetch depth needs to be high enough depending on the start date, though a depth of 1 is enough for only the latest signatures. Then its output file should be uploaded to the same rsync remote.

When needed I can of course create this file or make a workflow to do it for us, not a problem.

blattersturm:
I think it'd be best (also to serve as a dry run for the GH Actions workflow) to have a manually-triggered workflow for this, indeed.

I've merged the `compatgen` branch in `native-doc-tooling` by the way, https://github.com/citizenfx/native-doc-tooling/commit/528f3b05a9a3d92fda5896b5204670f8720817a1 is the current commit header for such.

thorium-cfx:
Sounds good, will get at it asap.

PR #863, opened 2023-02-09
Compatibility generation (manual only)

thorium-cfx:

Native compatibility generation for manual activation and testing.
* Does not activate on pull requests or any other than manual activation.

.ci/setup-builder
 * Use `$BASH_SOURCE` instead of `$0` allowing it to be included as well (`source`/`.`),
 * Shallow checkout of a specific commit from `native-doc-tooling`, no longer downloading its full history.

Notes:
 * use of rsync's NATIVE_DEPLOY_URL constants may not be correct as I don't know the format of its contents.

### DIFF
--- a/.ci/run-compat
+++ b/.ci/run-compat
@@ -1,0 +1,14 @@
+#!/bin/bash
+
+# configuration
+GENERATOR_PATH="$1/compatgen.js"
+OUTPUT_FILE=${2:-"./out/natives_global_client_compat.lua"}
+INPUT_FILE=${3:-"./in/natives_global_client_compat.lua"}
+IGNORE_MISSING_INPUT=${4:-false}
+FORCE_UPDATE=${5:-false}
+USE_HISTORY=${6:-false}
+DAYS_SINCE_DATE=${7:-2023-01-01} #YYYY-MM-DD
+VERSION=${8:-2.0}
+
+# execution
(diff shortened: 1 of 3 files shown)


blattersturm:
> use of rsync's NATIVE_DEPLOY_URL constants may not be correct

Good question - it seems one can't actually view the current value of a GH Actions secret but it *might* be unlucky and include the filename.

> Use $BASH_SOURCE instead of $0 

I was worried it was using a different shell in `#!` but it indeed appears to be `/bin/bash`.

thorium-cfx:
> Good question - it seems one can't actually view the current value of a GH Actions secret but it might be unlucky and include the filename.

For testing I used a format in which it ended with a folder path, so things like `$NATIVE_DEPLOY_URL/natives_global_client_compat.lua` worked nicely.
I did test everything with rsync with my own server, worked all fine, so I'd assume it holds the folder name as well as it transferred all files from the out folder to the remote folder.

Some outputs of some runs:
https://github.com/thorium-cfx/natives/actions/runs/4137585734/jobs/7152937828 (Manual)
https://github.com/thorium-cfx/natives/actions/runs/4127234844/jobs/7130189244 (Pull)

Running the manual workflow can give us the answer quickly with the unknown `NATIVE_DEPLOY_URL`.

PR #868, opened 2023-03-12
Update AttachCamToPedBone.md

matehun00:


4mmonium (commented):
This looks ok to me, and yeah, heading parameter was incorrect, nice catch!
`isRelative` seems correct after looking at [`AttachCamToEntity`](https://docs.fivem.net/natives/?_0xFEDB7D269E8C60E3) native method in disassembly and seeing it calls the exact same game method for argument/parameter `isRelative`.

You should explain what the parameters do though, what may seem obvious to you, may not be for other people 😃

PR #873, opened 2023-03-25
Compatibility generation (automatic, push)

thorium-cfx:

* Enables the compatgen on each push.
* Clone latest native-doc-tools instead of a specific commit.

### DIFF
--- a/.ci/run-deploy
+++ b/.ci/run-deploy
@@ -2,18 +2,23 @@
 
 set -eu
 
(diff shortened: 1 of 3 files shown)


blattersturm:
> No docs found to be built.

Can this failure be ignored?

thorium-cfx:
Re-ran it, seems so, they have the same result as the other non-native related PRs I did here.

PR #875, opened 2023-04-01
Update SetPedCanPlayAmbientIdles documentation

coalaura:
I was playing around with a variety of natives to try to disable the idle animation on player peds. This native did the trick, after experimenting with it, it turns out it only works if p1 and p2 are true. Later noticed it was not syncing between clients, so I had to have a loop in my script, calling it for each active player ped for it to work.


freedy69:
lol thank you for this i never understood how this native worked

TASK/SetPedCanPlayAmbientIdles.md
```diff
@@ -9,15 +9,16 @@ aliases: ["0x8FD89A6240813FD0"]
...
+while true do
+    SetPedCanPlayAmbientIdles(PlayerPedId(), true, true)
+    Citizen.Wait(0)
+end
 ```
 
 ## Parameters
 * **ped**: 
 * **p1**: 
 * **p2**:
```
PsychoShock, line 24: * **ped**: The ped index.
* **blockidleclips**: If true, blocks future idle clips from playing.
* **removeclip**: If true, will remove any current clip playing (will also block).


This is the missing params :D

coalaura: thank you! I'll update my PR

coalaura:

TASK/SetPedCanPlayAmbientIdles.md
coalaura: merged with master

coalaura:
Looking at it, i believe https://github.com/citizenfx/natives/commit/e478807fbc33fc940df2e35f7287614142fbbf8d is a better solution

PR #876, opened 2023-04-01
Update AddTextComponentSubstringKeyboardDisplay.md

coalaura:
I noticed when trying to draw a text including "<" it would draw the text up until that character and then stop. After experimenting with a bunch of different escape characters, i figured out you are able to use html entities to escape those characters.


coalaura:

PR #877, opened 2023-04-03
Update SetGameplayEntityHint.md

PeekerJ:


CAM/SetGameplayEntityHint.md
```diff
@@ -9,7 +9,11 @@ void SET_GAMEPLAY_ENTITY_HINT(Entity entity, float xOffset, float yOffset, float
 ```
 
 ```
-p6 & p7 - possibly length or time  
+p4 - boolean (true/false)
```
4mmonium, line 12: Hey there, thank you for your contribution, however:
Descriptions shouldn't be written inside code fences as they're used as a way to indicate that said documentation is the original, imported from NativeDB.

Here's the [contribution guide](https://github.com/citizenfx/natives/blob/master/.github/CONTRIBUTING.md).

Thank you once again!

4mmonium (changes requested):
Requested some changes, you can find them in the comment I made 🙂

PR #878, opened 2023-04-15
Update CAN_PHONE_BE_SEEN_ON_SCREEN description

tomilovsenya:
Add suggestion about using Global Variable 20365 which does what this native is supposed to.


4mmonium:
> Add suggestion about using Global Variable 20365 which does what this native is supposed to.

The native always returns 1 regardless. It takes an argument which is then hard-coded to `1`. It's also in the native imported description, I don't see how using a Global Variable would make a difference. Thanks for trying though!

```c
_DWORD *__fastcall n_CAN_PHONE_BE_SEEN_ON_SCREEN(_QWORD *a1)
{
  _DWORD *result; // rax

  result = (_DWORD *)*a1;
  *(_DWORD *)*a1 = 1;
  return result;
}
```

Disquse:
In FiveM there's no globals, so this addition makes no sense.

PR #885, opened 2023-06-04
Update RegisterPedheadshot.md

Mathu-lmn:
Added a code example and the return value


4mmonium (approved):
Tested the example using ped model `a_m_y_skater_01`.
Example seems similar to [this](https://github.com/citizenfx/natives/blob/master/HUD/EndTextCommandThefeedPostAward.md#examples).
LGTM, thank you for your contribution!

PR #886, opened 2023-06-09
Update GetPedEyeColor.md

coalaura:
Testing shows -1 has eye colors that don't exist in any of the other colors.

### DIFF
--- a/PED/GetPedEyeColor.md
+++ b/PED/GetPedEyeColor.md
@@ -15,7 +15,7 @@ A getter for [`_SET_PED_EYE_COLOR`](#_0x50B56988B170AFDF).
 * **ped**: The target ped
 
 ## Return value
-Returns ped's eye colour, or -1 if fails to get.
+Returns ped's eye colour (0 - 31), or -1 if the eye color is not set. If the eye color is not set it is inherited depending on the parents of the freemode model.
 
 ## Examples
 


PED/GetPedEyeColor.md
```diff
@@ -15,7 +15,7 @@ A getter for [`_SET_PED_EYE_COLOR`](#_0x50B56988B170AFDF).
 * **ped**: The target ped
 
 ## Return value
-Returns ped's eye colour, or -1 if fails to get.
+Returns ped's eye colour (0 - 31), or -1 if the eye color is not set. If the eye color is not set it is inherited depending on the parents of the freemode model.
```
4mmonium, line 18: I can't find proof about the eye color being inherited by the parents of the freemode model. I searched for offset references in the exe, but I couldn't get to that. All I know is that the `CPedHeadBlendData`  constructor sets the eye color to `-1` when it first gets called, hence why it's logical for it to return `-1` if not set.

I may run some tests later to find more.

coalaura: From testing around with it, there was a different eye color that was not in the list of regular eye colors and would change depending on the parents when set to -1

coalaura: In our server we had forcefully set the eye color after leaving it as -1 default for a while and we noticed that everyones eye color had changed to a color that was not replicable by any regular eye color in the list

coalaura:

AvarianKnight:
Closing this as the contents can't be validated.

PR #887, opened 2023-06-11
Update SetPedComponentVariation.md

KillStr3aK:
fixed bad link that referred to 404


4mmonium:
Perhaps it would be better to have this over on the [docs](https://docs.fivem.net) in the game references section and then add a link to it from here, rather than using links off-site which could go offline later in the future.

You could simply make a PR on [docs](https://github.com/citizenfx/fivem-docs) and then reference it from here once that PR is accepted.

Thank you for your contribution nonetheless 🙂

PR #888, opened 2023-06-13
Update SetPedConfigFlag.md

alberttheprince:
added note about flag 424, this has nothing to do with falling like an aircraft but puts a ped into a relaxed riding position at low speeds on motorcycles and in vehicles.


AvarianKnight:
Superseded by https://github.com/citizenfx/natives/pull/1082

PR #892, opened 2023-07-04
Correct the enum for SetPedCombatAttributes

4mmonium:
The previous enum declaration was indeed very wrong and was causing clients to crash.

### DIFF
--- a/PED/SetPedCombatAttributes.md
+++ b/PED/SetPedCombatAttributes.md
@@ -10,7 +10,9 @@ void SET_PED_COMBAT_ATTRIBUTES(Ped ped, int attributeIndex, BOOL enabled);
 
 
 These combat attributes seem to be the same as the BehaviourFlags from combatbehaviour.meta.
+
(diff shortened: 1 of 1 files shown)


gottfriedleibniz:
Given how long this has been broken (forever?), it may be worth having `NativeFixes` sanitize the parameter. I don't believe this enum has changed since 1604.

Looking at the other bitset-ish natives, this _seems_ like the only one that does not sanitize its length.

Disquse:
@gottfriedleibniz I made a `NativeFixes` code for this, gonna PR soon. And it actually changed somewhere between 1604 and 2060, so I made the check using parser. And yeah, that's indeed the only native with no "sanity checks", I was trying to find other potentially broken natives as well :)

gottfriedleibniz:
> And it actually changed somewhere between 1604 and 2060

Ah, `_0xEE129DBD = 90,` was added in 1868.

blattersturm (approved):

PR #894, opened 2023-07-06
Link patch

Mathu-lmn:
This PR fixes all the remaining old runtime.fivem.net links in the natives description and also renames some natives in the description as they got a name after their documentation creation

### DIFF
--- a/GRAPHICS/EnableAlienBloodVfx.md
+++ b/GRAPHICS/EnableAlienBloodVfx.md
@@ -9,10 +9,10 @@ aliases: ["0x9DCE1F0F78260875"]
 void ENABLE_ALIEN_BLOOD_VFX(BOOL toggle);
 ```
 
-Creates a motion-blur sort of effect, this native does not seem to work, however by using the [`START_SCREEN_EFFECT`](#_0x2206BF9A37B7F724) native with `"DrugsMichaelAliensFight"` as the effect parameter, you should be able to get the effect.
+Creates a motion-blur sort of effect, this native does not seem to work, however by using the [ANIMPOSTFX_PLAY](https://docs.fivem.net/natives/?_0x2206BF9A37B7F724) native with `"DrugsMichaelAliensFight"` as the effect parameter, you should be able to get the effect.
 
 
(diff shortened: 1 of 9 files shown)


AvarianKnight (changes requested):
Shouldn't these just use native hashes like `#_0xE35B38A27E8E7179` instead linking the website?

4mmonium:
> Shouldn't these just use native hashes like `#_0xE35B38A27E8E7179` instead linking the website?

Indeed, linking through the domain is not really a good idea, if the domain/subdomain ever changes for it, it will introduce issues, it's also not needed and it's better to have relative file paths.

GRAPHICS/EnableAlienBloodVfx.md
```diff
@@ -9,10 +9,10 @@ aliases: ["0x9DCE1F0F78260875"]
 void ENABLE_ALIEN_BLOOD_VFX(BOOL toggle);
 ```
 
-Creates a motion-blur sort of effect, this native does not seem to work, however by using the [`START_SCREEN_EFFECT`](#_0x2206BF9A37B7F724) native with `"DrugsMichaelAliensFight"` as the effect parameter, you should be able to get the effect.
+Creates a motion-blur sort of effect, this native does not seem to work, however by using the [ANIMPOSTFX_PLAY](https://docs.fivem.net/natives/?_0x2206BF9A37B7F724) native with `"DrugsMichaelAliensFight"` as the effect parameter, you should be able to get the effect.
```
4mmonium, line 12: Hey there, take a look at the following [line](https://github.com/citizenfx/natives/blob/b83b904e333fcd4ae80b5cf872899d8775d0be84/CAM/CreateCamWithParams.md?plain=1#L13) to see how this is done. It's better to not use domains and keep things relative, thank you for your contribution though!

4mmonium (changes requested):
Check out what I commented, thank you!

GRAPHICS/EnableAlienBloodVfx.md
Mathu-lmn: Will do thanks ! I did look in the natives website and many other natives were referenced using the full domain so that's why I did this

Mathu-lmn:
Closed and redone here : https://github.com/citizenfx/natives/pull/895

PR #896, opened 2023-07-08
Updated handful of ENTITY natives documentation

ook3D:
thanks for merging my last PR with other updated natives btw! 😄  i got encouraged to do some more. i added lua examples for each of the ones i updated and filled out the description and return values when applicable. let me know if theres anything that i need to fix. thanks in advance!

### DIFF
--- a/ENTITY/DeleteEntity.md
+++ b/ENTITY/DeleteEntity.md
@@ -13,3 +13,14 @@ Deletes the specified entity, and invalidates the passed handle (i.e. in/out arg
 
 ## Parameters
 - **entity**: The entity to delete.
+
+
+## Examples
+
+```lua
+local light = GetEntityModel(prop_streetlight_01)
+local lightEntity = CreateObject(lightEntity, 0, 0, 0, False, False, False)
+if DoesEntityExist(lightEntity) then
+    DeleteEntity(lightEntity)
+end
+```
--- a/ENTITY/DeleteEntity.md
+++ b/ENTITY/DeleteEntity.md
@@ -19,8 +19,7 @@ Deletes the specified entity, and invalidates the passed handle (i.e. in/out arg
 
 ```lua
 local light = GetEntityModel(prop_streetlight_01)
-local lightEntity = CreateObject(light, 0, 0, 0, False, False, False)
+local lightEntity = CreateObject(light, 0, 0, 0, false, false, false)
 if not DoesEntityExist(lightEntity) then return end
     DeleteEntity(lightEntity)
-end
 ```
--- a/ENTITY/DeleteEntity.md
+++ b/ENTITY/DeleteEntity.md
@@ -19,8 +19,8 @@ Deletes the specified entity, and invalidates the passed handle (i.e. in/out arg
 
 ```lua
 local light = GetEntityModel(prop_streetlight_01)
-local lightEntity = CreateObject(lightEntity, 0, 0, 0, False, False, False)
-if DoesEntityExist(lightEntity) then
+local lightEntity = CreateObject(light, 0, 0, 0, False, False, False)
+if not DoesEntityExist(lightEntity) then return end
     DeleteEntity(lightEntity)
 end
 ```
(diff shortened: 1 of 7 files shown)


ENTITY/DeleteEntity.md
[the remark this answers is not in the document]
ahcenezdh: you can do a JS / C# example too if you want

ENTITY/GetEntitySpeed.md
ahcenezdh, line 1: You can add a example to have the speed of a vehicle in km/h or miles/h based on the description of Return value

ENTITY/DeleteEntity.md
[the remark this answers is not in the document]
ook3D: I would do JS and C# examples, but im not confident they will be correct, and thanks ill tweak my example

ENTITY/GetEntitySpeed.md
ook3D: probably a good idea, thanks

ENTITY/DeleteEntity.md
```diff
@@ -13,3 +13,14 @@ Deletes the specified entity, and invalidates the passed handle (i.e. in/out arg
...
+
+
+## Examples
+
+```lua
+local light = GetEntityModel(prop_streetlight_01)
+local lightEntity = CreateObject(light, 0, 0, 0, False, False, False)
+if not DoesEntityExist(lightEntity) then return end
+    DeleteEntity(lightEntity)
+end
```
ahcenezdh, line 25: You forgot to remove a end line 25

ENTITY/DeleteEntity.md
```diff
@@ -19,8 +19,8 @@ Deletes the specified entity, and invalidates the passed handle (i.e. in/out arg
 
 ```lua
 local light = GetEntityModel(prop_streetlight_01)
-local lightEntity = CreateObject(lightEntity, 0, 0, 0, False, False, False)
-if DoesEntityExist(lightEntity) then
+local lightEntity = CreateObject(light, 0, 0, 0, False, False, False)
+if not DoesEntityExist(lightEntity) then return end
     DeleteEntity(lightEntity)
 end
```
ahcenezdh, line 25: You forgot to remove the end line 25

ENTITY/DeleteEntity.md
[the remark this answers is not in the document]
4mmonium: > instead of doing if DoesEntityExist(....), you can do
> 
> ```lua
> if not DoesEntityExist(lightEntity) then return end
> DeleteEntity(lightEntity)
> ```
> 
> and where `prop_streetlight_01` is defined here?

I don't really agree with putting everything in a single line for code examples, it's somewhat bad for readability. While we may understand it, new users that haven't been fully exposed to the language most likely won't.

ENTITY/DeleteEntity.md
```diff
@@ -13,3 +13,14 @@ Deletes the specified entity, and invalidates the passed handle (i.e. in/out arg
 
 ## Parameters
 - **entity**: The entity to delete.
+
+
+## Examples
+
+```lua
+local light = GetEntityModel(prop_streetlight_01)
+local lightEntity = CreateObject(light, 0, 0, 0, False, False, False)
```
4mmonium, line 22: Why are the booleans upper-cased? 🙂

4mmonium (changes requested):
Overall this seems fine, except for the one line conditional statements. While concise code can sometimes be desirable, I believe it may be better to prioritize readability and maintainability when writing code examples.

If you could correct that, it would be great.
Thank you for your contribution nonetheless! 🙂

ENTITY/GetEntityType.md
```diff
@@ -8,15 +8,22 @@ ns: ENTITY
...
 3 = object
-```
 
-## Parameters
-* **entity**: 
+## Examples
 
-## Return value
+```lua
+local light = GetEntityModel(prop_streetlight_01)
```
4mmonium, line 27: You should explain how to get the entity handle here before you pass it, and the `lightType` variable name makes me think there's multiple types of lights😅

Also not sure why you're getting the model, `GetEntityType` takes an entity handle.

So I would write this the following way:

```lua
local entityType = GetEntityType(PlayerPedId()) -- entityType should be '4'
```

PR #899, opened 2023-07-25
fix: Flag for `StartShapeTestLosProbe`

JajarGaming:
~~Found that the flag 511 is used the most in decompiled scripts~~

Fixed a flag...

### DIFF
--- a/SHAPETEST/StartShapeTestLosProbe.md
+++ b/SHAPETEST/StartShapeTestLosProbe.md
@@ -17,11 +17,12 @@ enum TraceFlags
   None = 0,
   IntersectWorld = 1,
   IntersectVehicles = 2,
-  IntersectPedsSimpleCollision = 4,
-  IntersectPeds = 8,
+  IntersectPeds = 4,
+  IntersectRagdoll = 8,
   IntersectObjects = 16,
   IntersectWater = 32,
-  Unknown = 128,
+  IntersectGlass = 64,
+  IntersectRiver = 128,
   IntersectFoliage = 256,
   IntersectEverything = -1
 }
--- a/SHAPETEST/StartShapeTestLosProbe.md
+++ b/SHAPETEST/StartShapeTestLosProbe.md
@@ -23,7 +23,6 @@ enum TraceFlags
   IntersectWater = 32,
   Unknown = 128,
   IntersectFoliage = 256,
-  IntersectEverything = 511
   IntersectEverything = -1
 }
 ```
--- a/SHAPETEST/StartShapeTestLosProbe.md
+++ b/SHAPETEST/StartShapeTestLosProbe.md
@@ -23,8 +23,8 @@ enum TraceFlags
   IntersectWater = 32,
   Unknown = 128,
   IntersectFoliage = 256,
-  Unkown = 511, (used most)
-  IntersectEverything = 4294967295
+  IntersectEverything = 511
+  IntersectEverything = -1
 }
 ```
 NOTE: Raycasts that intersect with mission_entites (flag = 2) has limited range and will not register for far away entites. The range seems to be about 30 metres.  
--- a/SHAPETEST/StartShapeTestLosProbe.md
+++ b/SHAPETEST/StartShapeTestLosProbe.md
@@ -24,7 +24,7 @@ enum TraceFlags
   IntersectGlass = 64,
   IntersectRiver = 128,
   IntersectFoliage = 256,
-  IntersectEverything = -1
+  IntersectAll = 511
 }
 ```
 NOTE: Raycasts that intersect with mission_entites (flag = 2) has limited range and will not register for far away entites. The range seems to be about 30 metres.  
--- a/SHAPETEST/StartShapeTestLosProbe.md
+++ b/SHAPETEST/StartShapeTestLosProbe.md
@@ -23,6 +23,7 @@ enum TraceFlags
   IntersectWater = 32,
   Unknown = 128,
   IntersectFoliage = 256,
+  Unkown = 511, (used most)
   IntersectEverything = 4294967295
 }
 ```
--- a/SHAPETEST/StartShapeTestLosProbe.md
+++ b/SHAPETEST/StartShapeTestLosProbe.md
@@ -6,7 +6,7 @@ aliases: ["0x7EE9F5D83DD4F90E"]
 
 ```c
 // 0x7EE9F5D83DD4F90E 0xEFAF4BA6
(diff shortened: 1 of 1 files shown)


SHAPETEST/StartShapeTestLosProbe.md
```diff
@@ -23,6 +23,7 @@ enum TraceFlags
   IntersectWater = 32,
   Unknown = 128,
   IntersectFoliage = 256,
+  Unkown = 511, (used most)
```
FlawwsX, line 26: Unk**n**own is the correct spelling

Disquse: `511` is basically "everything" (`1+ 2 + 4 + 8 + 16 + 32 + 64 + 128 + 256`)

JajarGaming: > Unk**n**own is the correct spelling

Shit just noticed

JajarGaming: > `511` is basically "everything" (`1 + 2 + 4 + 8 + 16 + 32 + 64 + 128 + 256`), it's not a single flag

Alright, I'll make the change

SHAPETEST/StartShapeTestLosProbe.md
```diff
@@ -23,6 +23,7 @@ enum TraceFlags
...
   IntersectEverything = 4294967295
```
4mmonium, line 27: I believe this should be `-1`, but it incorrectly got displayed as `4294967295` (`0xFFFFFFFF`). I have seen that happen multiple times in other unrelated cases. Maybe a test with the value being `-1` rather than  (`4294967295`) should be done in order to verify.

JajarGaming: I'm working on raycasting so I'll check it out

JajarGaming:
Going to make the changes and tests sometime this week

JajarGaming:
Done

SHAPETEST/StartShapeTestLosProbe.md
JajarGaming: Can confirm now, -1 intersects everything

SHAPETEST/StartShapeTestLosProbe.md
```diff
@@ -23,8 +23,8 @@ enum TraceFlags
   IntersectWater = 32,
   Unknown = 128,
   IntersectFoliage = 256,
-  Unkown = 511, (used most)
-  IntersectEverything = 4294967295
+  IntersectEverything = 511
+  IntersectEverything = -1
```
4mmonium, line 27: Why would you define `IntersectEverything` twice, that's wrong 🙂

JajarGaming: Aren't they the same basically?

4mmonium: > Aren't they the same basically?

No, that's wrong. Flags are usually represented using powers of two. Each flag is assigned a unique value that is a power of two (1, 2, 4, 8, 16, etc...), and we combine these values using bitwise operations to represent different sets of options.

Now, the part that doesn't make sense is defining 511 (for `IntersectEverything`) as a flag in the enum. In proper flag representation, we should stick to powers of two. 511 is not a power of two; it's `2^9 - 1`. That's why you see it displayed as `4294967295 (0xFFFFFFFF)` also **(known as -1, which is the correct representation here)**, which is the maximum unsigned value for a 32-bit integer.

To avoid confusion and ensure consistency, it's best to correct the enum by using `-1` for `IntersectEverything`, as that's a common convention to represent all available options. This way, the enum will work correctly with bitwise operations, and we won't encounter display errors or unexpected behavior.

In other words, remove `IntersectEverything = 511` from the enum.
More about power of two [here](https://en.wikipedia.org/wiki/Power_of_two) (Computer Science section) and [here](https://en.wikipedia.org/wiki/Bitwise_operation) for bit-wise operations.

Essentially, `TraceFlags` should be this:

```c
enum TraceFlags
{
    None = 0,
    IntersectWorld = 1,                 // 2^0
    IntersectVehicles = 2,              // 2^1
    IntersectPedsSimpleCollision = 4,   // 2^2
    IntersectPeds = 8,                  // 2^3
    IntersectObjects = 16,              // 2^4
    IntersectWater = 32,                // 2^5
    Unknown = 64,                       // 2^6
    IntersectFoliage = 128,             // 2^7
    IntersectEverything = -1,           // All flags combined using bitwise OR (Power of 2 combination)
}
```

gottfriedleibniz: For reference, `0x140D7CF2B/2904` maps these intersection flags to [composite flags](https://github.com/dexyfex/CodeWalker/blob/master/CodeWalker.Core/GameFiles/Resources/Bounds.cs#L4239)[^1] 

[^1]: Well, what is labelled in my IDB as `EBoundCompositeFlags` anyway.

SHAPETEST/StartShapeTestLosProbe.md
```diff
@@ -23,7 +23,7 @@ enum TraceFlags
   IntersectWater = 32,
   Unknown = 128,
```
PsychoShock, line 24: Unknow value here is the river.

PsychoShock:
Still think that correct flag is `511` and not `-1` from what I can see.

JajarGaming:
> Still think that correct flag is `511` and not `-1` from what I can see.

And yeah after further research I've noticed that as well

PsychoShock (approved):

AvarianKnight:
You should probably read @4mmonium's [previous comment](https://github.com/citizenfx/natives/pull/899#discussion_r1277044463).

SHAPETEST/StartShapeTestLosProbe.md
PsychoShock: `511` is still a valid flag. Probably remove it but make a note that this flag exist.

SHAPETEST/StartShapeTestLosProbe.md
```diff
@@ -17,13 +17,14 @@ enum TraceFlags
...
+  IntersectPeds = 4,
+  IntersectRagdolls = 8,
   IntersectObjects = 16,
   IntersectWater = 32,
-  Unknown = 128,
+  IntersectGlass = 64,
+  IntersectRiver = 128,
   IntersectFoliage = 256,
-  IntersectEverything = 4294967295
+  IntersectAll = 511
```
AvarianKnight, line 27:
```suggestion
  IntersectEverything = -1
```

AvarianKnight:
You didn't rename `p8` to `options` in the function declaration

PR #904, opened 2023-07-30
Update SetParticleFxLoopedColour.md

XenoS-ITA:
added a more precise description for color ranges and also a tip on how to convert an rgb to the necessary range.
for the external link it would be better to find a method to make it "permanent" i read the advice to add it to the fivem documentation but idk where to put it

### DIFF
--- a/GRAPHICS/SetParticleFxLoopedColour.md
+++ b/GRAPHICS/SetParticleFxLoopedColour.md
@@ -17,4 +17,4 @@ You can use the [inverse lerp](https://www.gamedev.net/articles/programming/gene
 * **r**: range from 0.0 to 1.0
 * **g**: range from 0.0 to 1.0
 * **b**: range from 0.0 to 1.0
-* **bLocalOnly**: Defines whether the effect should be modified over the network or not. Set this to false if you want to apply the effect over the network.
+* **bLocalOnly**: False by default. Defines whether the effect should be sent and modified over the network or not. Set this to `true` if you don't want to send the effect over the network.
--- a/GRAPHICS/SetParticleFxLoopedColour.md
+++ b/GRAPHICS/SetParticleFxLoopedColour.md
@@ -8,15 +8,15 @@ ns: GRAPHICS
 void SET_PARTICLE_FX_LOOPED_COLOUR(int ptfxHandle, float r, float g, float b, BOOL p4);
 ```
 
-```
-only works on some fx's  
-p4 = 0  
-```
-
 ## Parameters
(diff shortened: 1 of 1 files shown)


4mmonium:
This seems to be correct, `r`, `g`, `b` values need to range from `0.0` to `1.0` given how they are multiplied by `255.0` in game code.

i.e.

```
v11[151] = (r * 255.0); // red
```

`p4` apparently needs to be `false` for this block of code to even execute.

XenoS-ITA:
probably then the `p4` is the parameter that decides whether the values are "normalized" so `0.0` to `1.0` or whether they are normal `rgb` values

GRAPHICS/SetParticleFxLoopedColour.md
```diff
@@ -8,15 +8,15 @@ ns: GRAPHICS
...
+* **r**: range from 0.0 to 1.0
+* **g**: range from 0.0 to 1.0
+* **b**: range from 0.0 to 1.0
 * **p4**: 
 
+
+only works on some fx's  
+p4 = 0
+
+You can use the [inverse lerp](https://www.gamedev.net/articles/programming/general-and-gameplay-programming/inverse-lerp-a-super-useful-yet-often-overlooked-function-r5230/) method to normalize in a range from 0.0 to 1.0 an rgb
```
4mmonium, line 22: I just realized, all this needs to be above the the `## Parameters` section. It would also be nice if you clarified on which fx's this works on and elaborated on why `p4` needs to be false, rather than just `p4 = 0` 🙂

4mmonium (changes requested):

GRAPHICS/SetParticleFxLoopedColour.md
```diff
@@ -8,15 +8,15 @@ ns: GRAPHICS
...
-
 ## Parameters
 * **ptfxHandle**: 
-* **r**: 
-* **g**: 
-* **b**: 
+* **r**: range from 0.0 to 1.0
+* **g**: range from 0.0 to 1.0
+* **b**: range from 0.0 to 1.0
 * **p4**:
```
PsychoShock, line 16: Missing param is if is local only. False by default.

PsychoShock: It will also answer @4mmonium comments on
> It would also be nice if you clarified on which fx's this works on and elaborated on why p4 needs to be false, rather than just p4 = 0 🙂

4mmonium (approved):
Made some changes, LGTM, merging. Thanks!

PR #906, opened 2023-08-05
Cleanup RECORDING natives

niekschoemaker:
Minor cleanup of documentation


AvarianKnight:
It might be useful to point to other natives that's used with this native, also everything in the description should just be in the return.

4mmonium (approved):
LGTM, merging, thank you!

PR #908, opened 2023-08-10
Improve DistantCopCarSirens documentation

barbeque-squared:
The distant sirens when _any_ player had their emergency lights on (regardless of whether the vehicle actually has a siren) are annoying. At first I was unsure of whether this native would fix that problem (the description reads like it's just on/off) but after some testing it turns out:
- the default is `true`
- when set to `true`, it will only actually play the sound when _any_ player has emergency lights on. If no-one has emergency lights on, the sound isn't played.
- when toggling it off while the sound is playing, the distant sirens actually fade out, and vice versa.

Usecase: we have a race grid of about 30 cars, plus 1 safety car which has emergency lights (only lights, no actual siren). All players _except_ the safety car player and the vehicles very near to it would hear constant or intermittent distant sirens. Setting it to false has resolved this issue.

NB: I'm not sure if this also affects NPC sirens. I only tested it with actual players.
NB2: feel free to add more commits before merging if you want to reword things.

### DIFF
--- a/AUDIO/DistantCopCarSirens.md
+++ b/AUDIO/DistantCopCarSirens.md
@@ -10,8 +10,9 @@ void DISTANT_COP_CAR_SIRENS(BOOL value);
 ```
 
 ```
-If value is set to true, and ambient siren sound will be played.
+If value is set to true, an ambient siren sound will be played if another player is in a vehicle that has the emergency lights on, even if the vehicle has no actual siren.
 Appears to enable/disable an audio flag.
+Defaults to true.
 ```
 
 ## Parameters


AUDIO/DistantCopCarSirens.md
```diff
@@ -10,8 +10,9 @@ void DISTANT_COP_CAR_SIRENS(BOOL value);
 ```
 
 ```
-If value is set to true, and ambient siren sound will be played.
+If value is set to true, an ambient siren sound will be played if another player is in a vehicle that has the emergency lights on, even if the vehicle has no actual siren.
 Appears to enable/disable an audio flag.
+Defaults to true.
```
PsychoShock, line 15: You could just use `Plays/stop a fake distant siren sound.` since this native aims to play a "fake" sound. 

Also default value seems to be `false` since this is only toggle when needed. You can modify the param name/description to toggle since it's more about `true` or `false` to player or stop the sound.

PsychoShock (changes requested):

AUDIO/DistantCopCarSirens.md
barbeque-squared: Unless something changed since I originally made the PR (sorry I already rebased it before commenting this, I'm just too used to it), it's **not** that setting `true` starts playing a fake siren regardless of what's happening around you; if you're on an empty server (without traffic) you won't ever heard it even if this is set to `true`.

It's possible that your statement is true if you're currently in the city (maybe it also controls _those_ sirens?) but I tested this near the observatory and it will probably also work like I described on Chiliad. That's why I found the existing documentation so confusing.

PsychoShock: Do you have a code with what you tested? A repo? I feel that your vehicle engine was probably shut down and consequently stopped the native to play. This native is only playing a fake distant siren sound, usually used for people to make vehicles like "cops". It could even be called PlayDistFakeCopCarSirens.

barbeque-squared: Once I have time (February, probably) I'll figure out some reproducable way of doing this.

barbeque-squared: Finally had some time, I can confirm that it is indeed (still? again?) playing distant sirens when another player is between "not super close" and "not super far away", making this PR invalid.
It's possible it still does _something_ but at least for me it's not consistent and not worth looking further into.

PR #912, opened 2023-08-27
Explain TOGGLE_PAUSED_RENDERPHASES

DatGuy1:
Also rename variable

### DIFF
--- a/GRAPHICS/TogglePausedRenderphases.md
+++ b/GRAPHICS/TogglePausedRenderphases.md
@@ -6,10 +6,10 @@ aliases: ["_ENABLE_GAMEPLAY_CAM","_SET_FROZEN_RENDERING_DISABLED"]
 
 ```c
 // 0xDFC252D8A3E15AB7 0x30ADE541
-void TOGGLE_PAUSED_RENDERPHASES(BOOL toggle);
+void TOGGLE_PAUSED_RENDERPHASES(BOOL rendering);
 ```
 
+Freezes the screen in its current state.
 
 ## Parameters
-* **toggle**: 
-
+* **rendering**: Whether to actively render the camera
--- a/GRAPHICS/TogglePausedRenderphases.md
+++ b/GRAPHICS/TogglePausedRenderphases.md
@@ -12,4 +12,4 @@ void TOGGLE_PAUSED_RENDERPHASES(BOOL toggle);
 Switches the rendering display to exclude everything except PostFX, resulting in a frozen screen before the UI pass.
 
 ## Parameters
-* **toggle**: Setting it to `false` will halt render updates as mentioned in the description, until it's set back to `true`.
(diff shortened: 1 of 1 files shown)


Z3rio (changes requested):
Is the change from `toggle` to `rendering` really needed?
In my opinion, `toggle` does sound better.

Other than that, the descriptions looked fairly good to me.
the param description could probably be a bit better/easier to understand though, such as: `Whether to free the camera or not`

4mmonium:
I think the original parameter name is fine, not sure why it would need a name change.
The original description needs improvements though. This function pauses game rendering when toggled off and resumes when toggled on (very similar to what the game does in single player when you press ESC to open the pause menu when the world is loaded).

DatGuy1:
I renamed it from 'toggle' to rendering because the native name together with the argument name doesn't make it clear which status it toggles to, which is the primary function of an argument name. Does it toggle rendering on? Does it toggle paused rendering on, meaning it doesn't render?

Z3rio:
> I renamed it from 'toggle' to rendering because the native name together with the argument name doesn't make it clear which status it toggles to, which is the primary function of an argument name. Does it toggle rendering on? Does it toggle paused rendering on, meaning it doesn't render?

The variable/param name "toggle" sounds a whole lot better though, and is uniform with the rest of the documentation.
And the issue you are explaining with that name can be solved with simply adding descriptions for the native & param, which you seem to be doing in this PR.

GRAPHICS/TogglePausedRenderphases.md
```diff
@@ -6,10 +6,10 @@ aliases: ["_ENABLE_GAMEPLAY_CAM","_SET_FROZEN_RENDERING_DISABLED"]
 
 ```c
 // 0xDFC252D8A3E15AB7 0x30ADE541
-void TOGGLE_PAUSED_RENDERPHASES(BOOL toggle);
+void TOGGLE_PAUSED_RENDERPHASES(BOOL rendering);
 ```
 
+Freezes the screen in its current state.
```
PsychoShock, line 12: Better description would be `Switches the rendering display to exclude everything except PostFX, resulting in a frozen screen before the UI pass.`

PsychoShock (changes requested):

GRAPHICS/TogglePausedRenderphases.md
```diff
@@ -6,10 +6,10 @@ aliases: ["_ENABLE_GAMEPLAY_CAM","_SET_FROZEN_RENDERING_DISABLED"]
...
 
 ## Parameters
-* **toggle**: 
-
+* **rendering**: Whether to actively render the camera
```
PsychoShock, line 15: This should be `toggle` since it's a "toggle" native.

4mmonium (approved):
LGTM, merging, thank you! 😊

Issue #910, opened 2023-08-15
CreateMissionTrain changed in 2802


TonybynMp4:
The docs states that the metro variation is 24 before 2372 and 25 after, it is also 26 after 2802

4mmonium:
Resolved in #915 
Thank you for reporting this!

PR #915, opened 2023-09-03
CreateMissionTrain: Fix `Double metro train` variation

Z3rio:
This PR aims to resolve issue #910.

The issue refers to the train variant `Double metro train`, who's number has been changed from `25` to `26`.
Please note that this PR is only based off of the info in the issue, it has not been tested in-game yet, so please refrain from merging before it has actually been verified that such change is correct.


4mmonium (approved):
This seems correct, thank you for your contribution. I tested 2802 using fxDK and given the code example provided in this native, it indeed seems that there's a new train variation, thus causing the metro variation to shift from id 25 to 26 in 2802.

PR #916, opened 2023-09-14
Cleanup Pass 1

LWSS:
manual pass mainly focused on param types

basic test done with native-doc-tooling


AvarianKnight:
How is this going to end up working for over all compatibility?

LWSS:
> How is this going to end up working for over all compatibility?

the older game versions are going to be removed eventually

AvarianKnight:
> > How is this going to end up working for over all compatibility?
> 
> the older game versions are going to be removed eventually

But that doesn't solve the problem that changing the ABI will cause mono v1 to break

LWSS:
> > > How is this going to end up working for over all compatibility?
> > 
> > 
> > the older game versions are going to be removed eventually
> 
> But that doesn't solve the problem that changing the ABI will cause mono v1 to break

Thanks for pointing that out

Yeah this will have to wait until v2, or if we lock v1's natives state

AvarianKnight:
Both

ahcenezdh:
So why not using cs_type?

AvarianKnight:
Because a lot of natives are old and have gotten new arguments over time. Adding new arguments would break ABI compatibility with existing C# resources causing them to fail to run. 

This is fixed with mono_v2, but it looks like its going to be a long time until mono_v2 is ready

PR #930, opened 2023-11-07
feat(natives/vehicle): Update SetConvertibleRoof

ahcenezdh:
Here is a more detailed documentation for SetConvertibleRoof.

### DIFF
--- a/VEHICLE/SetConvertibleRoof.md
+++ b/VEHICLE/SetConvertibleRoof.md
@@ -5,7 +5,7 @@ ns: VEHICLE
 
 ```c
 // 0xF39C4F538B5124C2 0xC87B6A51
-void SET_CONVERTIBLE_ROOF(Vehicle vehicle, BOOL roofState);
+void SET_CONVERTIBLE_ROOF(Vehicle vehicle, BOOL toggle);
 ```
 This allows for the vehicle's roof to be put on when set to true, and removed when set to false, provided that the vehicle has a version with a roof and a version without a roof.
 
@@ -23,7 +23,7 @@ This allows for the vehicle's roof to be put on when set to true, and removed wh
 
 ## Parameters
 * **vehicle**: The vehicle to which the roof state will be applied.
-* **roofState**: Boolean value where `true` adds the roof and `false` removes it, applicable only if the vehicle has a version with and without a roof.
+* **toggle**: Boolean value where `true` adds the roof and `false` removes it, applicable only if the vehicle has a version with and without a roof.
 
 ## Examples
 ```lua
--- a/VEHICLE/SetConvertibleRoof.md
+++ b/VEHICLE/SetConvertibleRoof.md
@@ -5,11 +5,42 @@ ns: VEHICLE
 
 ```c
 // 0xF39C4F538B5124C2 0xC87B6A51
-void SET_CONVERTIBLE_ROOF(Vehicle vehicle, BOOL p1);
+void SET_CONVERTIBLE_ROOF(Vehicle vehicle, BOOL roofState);
 ```
+This allows for the vehicle's roof to be put on when set to true, and removed when set to false, provided that the vehicle has a version with a roof and a version without a roof.
+
+#### Vehicles with both roofed and roofless versions (others may exist; this list is compiled from decompiled scripts).
+
+* chino
+* voltic
+* buccaneer
+* buccaneer2
+* chino2
+* faction
+* faction2
+* mamba
 
 
 ## Parameters
-* **vehicle**: 
-* **p1**: 
+* **vehicle**: The vehicle to which the roof state will be applied.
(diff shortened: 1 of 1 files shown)


ahcenezdh:
Can i change the argument p1 and put something like roofState instead?

4mmonium:
> Can i change the argument p1 and put something like roofState instead?

Yes you can, maybe you can name it `toggle` as well?

VEHICLE/SetConvertibleRoof.md
```diff
@@ -7,9 +7,39 @@ ns: VEHICLE
 // 0xF39C4F538B5124C2 0xC87B6A51
 void SET_CONVERTIBLE_ROOF(Vehicle vehicle, BOOL p1);
 ```
-
+# This allows for the vehicle's roof to be put on when set to true, and removed when set to false, provided that the vehicle has a version with a roof and a version without a roof.
```
4mmonium, line 10: Remove the `#`.

VEHICLE/SetConvertibleRoof.md
```diff
@@ -7,9 +7,39 @@ ns: VEHICLE
...
 
 ## Parameters
-* **vehicle**: 
-* **p1**: 
+* **vehicle**: The vehicle to which the roof state will be applied.
+* **p1**: Boolean value where `true` adds the roof and `false` removes it, applicable only if the vehicle has a version with and without a roof.
+
+## Vehicles with both roofed and roofless versions (others may exist; this list is compiled from decompiled scripts).
```
4mmonium, line 16: No need for the double hash (`##`), maybe you can do this:

```
#### Vehicles with both roofed and roofless versions (others may exist; this list is compiled from decompiled scripts).

* Chino
* voltic
* buccaneer
* buccaneer2
* chino2
* faction
* faction2
* mamba
```

VEHICLE/SetConvertibleRoof.md
```diff
@@ -7,9 +7,39 @@ ns: VEHICLE
...
 void SET_CONVERTIBLE_ROOF(Vehicle vehicle, BOOL p1);
 ```
-
+# This allows for the vehicle's roof to be put on when set to true, and removed when set to false, provided that the vehicle has a version with a roof and a version without a roof.
 
 ## Parameters
-* **vehicle**: 
-* **p1**: 
+* **vehicle**: The vehicle to which the roof state will be applied.
+* **p1**: Boolean value where `true` adds the roof and `false` removes it, applicable only if the vehicle has a version with and without a roof.
```
4mmonium, line 14: Maybe rename `p1` to `toggle`, or `roofState`, but I believe this toggles a roof on/off.

VEHICLE/SetConvertibleRoof.md
```diff
@@ -7,9 +7,39 @@ ns: VEHICLE
 // 0xF39C4F538B5124C2 0xC87B6A51
 void SET_CONVERTIBLE_ROOF(Vehicle vehicle, BOOL p1);
```
4mmonium, line 8: Maybe rename `p1` to `toggle`, or `roofState`, but I believe this toggles a roof on/off.

4mmonium (changes requested):
Hey there, couple changes needed, good otherwise! Thank you for your contribution 👍

ahcenezdh:
Should be good @4mmonium !

VEHICLE/SetConvertibleRoof.md
```diff
@@ -5,11 +5,41 @@ ns: VEHICLE
...
+#### Vehicles with both roofed and roofless versions (others may exist; this list is compiled from decompiled scripts).
+
+* chino
+* voltic
+* buccaneer
+* buccaneer2
+* chino2
+* faction
+* faction2
+* mamba
```
4mmonium, line 25: All this needs to be moved above the `## Parameters` section (under 'This allows for...'), thank you.

VEHICLE/SetConvertibleRoof.md
```diff
@@ -5,11 +5,41 @@ ns: VEHICLE
...
+
+## Examples
+```lua
+-- In this case we are removing the roof from the vehicle.
+local vehicle = GetVehiclePedIsIn(PlayerPedId(), false)
+SetConvertibleRoof(vehicle, false)
+```
+
+```javascript
```
4mmonium, line 34: Whoops, I must have missed this, replace with `` ```js ``, not javascript, thanks!

ahcenezdh: No problem, should be good

VEHICLE/SetConvertibleRoof.md
4mmonium: You missed this in your last commit  :-)

ahcenezdh: Oh sorry, it's good like this?

PR #931, opened 2023-11-07
feat(natives/vehicles): Update natives about hover mode (deluxo/opres…

ahcenezdh:
The code provided in the examples has only been tested in Lua; there may be errors


ahcenezdh:
So i don't know why the checks are failing :(

VEHICLE/SetSpecialFlightModeAllowed.md
```diff
@@ -0,0 +1,39 @@
...
+aliases: ["0xF1211889DF15A763", "_SET_VEHICLE_HOVER_TRANSFORM_ENABLED"]
+---
+## SET_SPECIAL_FLIGHT_MODE_ALLOWED
+
+```c
+// 0xF1211889DF15A763
+void SET_SPECIAL_FLIGHT_MODE_ALLOWED(Vehicle vehicle, BOOL toggle);
+```
+
+## Allows locking the hover/non-hover mode of a vehicle, such as the flying mode of the `Deluxo`. In the decompiled scripts, this native is used on `oppressor2` but couldn't get it to work on it.
```
4mmonium, line 12: Remove the double hash (`##`), thanks!

4mmonium (changes requested):
Hey, thank you for your contribution. I reviewed this PR, it's OK for the most part with a couple issues here and there, a couple changes and it's good to go 😀

We have a contribution guide if you want to check it out ([CONTRIBUTING.md](https://github.com/citizenfx/natives/blob/master/.github/CONTRIBUTING.md)).

ahcenezdh:
@4mmonium I made all the requested changes

4mmonium:
@thorium-cfx This changes a couple native file names and adds aliases (I'm pretty sure it's fine), but I assigned this to you to make sure, thanks!

4mmonium (commented):
I made some changes to your PR, check the commits, they mostly involved the double hash usage that is reserved for example and parameters (custom headers using double slashes are not allowed, one may use triple hashes instead). I cannot approve this yet, but good job nonetheless! 😊

thorium-cfx:
> @thorium-cfx This changes a couple native file names and adds aliases (I'm pretty sure it's fine), but I assigned this to you to make sure, thanks!

Aliased: check.
Same parameters: check.
LGTM

Would be even nicer if these files can be marked as moved, probably need to do it with `git` directly in this case.

4mmonium (approved):

PR #935, opened 2023-11-23
feat(natives/streaming): examples and updated natives for switch transition

ahcenezdh:
1. **SetCloudHatOpacity:**
   - Updated the native `SetCloudHatOpacity`.
   - Added a new example showcasing its usage.

2. **SwitchInPlayer and SwitchOutPlayer:**
   - Made updates to `SwitchInPlayer` and `SwitchOutPlayer` natives.
   - Ensured consistency in the description and examples.

3. **GetPlayerSwitchState:**
   - Updated the description and examples for the `GetPlayerSwitchState` native.


AvarianKnight:
LGTM

MISC/SetCloudsAlpha.md
```diff
@@ -0,0 +1,45 @@
...
+## SET_CLOUDS_ALPHA
+
+```c
+// 0xF36199225D6D8C86
+void SET_CLOUDS_ALPHA(float opacity);
+```
+
+Allows modification of the cloud intensity when the player is in a switch state [`IS_PLAYER_SWITCH_IN_PROGRESS`](#_0xD9D2CFFF49FAB35F).
+
+You can set the player in a switch state with [`SWITCH_TO_MULTI_FIRSTPART`](#_0xAAB3200ED59016BC).
```
4mmonium, line 14: I'm afraid this allows the modification of the cloud's alpha at any given time, the player doesn't need to be switching states or anything alike.

I was coincidentally researching this and the cloud opacity is actually a global float that can be set via `SET_CLOUDS_ALPHA`.  That float is then read via a `CRenderPhaseDrawScene` method where it's added to a draw command data structure.

This is how I tested it:

```lua
RegisterCommand('cloudopacity', function(source, args)
    SetWeatherTypeNowPersist('CLOUDS')
    SetCloudHatOpacity(tonumber(args[1])) -- Original native name
end, false)
```

STREAMING/GetPlayerSwitchState.md
```diff
@@ -1,5 +1,6 @@
 ---
 ns: STREAMING
+aliases: ["0x470555300D10B2A5"]
```
4mmonium, line 3: You should only need to add the alias if you actually changed the function name. Aliases are mainly there to preserve compatibility with game modes that maintain older native names.

STREAMING/SwitchToMultiSecondpart.md
```diff
@@ -0,0 +1,48 @@
...
+RegisterCommand("switchPlayer", function()
+    if IsPlayerSwitchInProgress() then return end
+    local ped = PlayerPedId()
+    SwitchToMultiFirstPart(ped, 0, 1)
+    Citizen.Wait(5000)
+    SwitchToMultiSecondPart(ped)
+end, false)
+```
+
+```javascript
```
4mmonium, line 28: I'm not sure whether `` ```javascript `` is supported on docs (couldn't find any matches), better use the more prominent `` ```js `` version of it.

STREAMING/SwitchToMultiFirstpart.md
```diff
@@ -0,0 +1,51 @@
...
+```lua
+-- Check if the player is in a Switch "state"
+if not IsPlayerSwitchInProgress() then
+    -- If the player is not already in a Switch state, initiate a Switch
+    SwitchToMultiFirstPart(PlayerPedId(), 0, 1)
+    -- In this case, switchType is set to 1, which means "3 steps out from ped"
+end
+```
+
+```javascript
```
4mmonium, line 33: I'm not sure whether `` ```javascript `` is supported on docs (couldn't find any matches), better use the more prominent `` ```js `` version of it.

STREAMING/SwitchToMultiFirstpart.md
```diff
@@ -0,0 +1,51 @@
...
+// Check if the player is in a Switch "state"
+if (!IsPlayerSwitchInProgress()) {
+    // If the player is not already in a Switch state, initiate a Switch
+    SwitchToMultiFirstPart(PlayerPedId(), 0, 1);
+    // In this case, switchType is set to 1, which means "3 steps out from ped" according to the documentation
+}
+```
+
+```csharp
```
4mmonium, line 42: `` ```cs `` is more common for docs, but this may work.

STREAMING/SwitchToMultiSecondpart.md
```diff
@@ -0,0 +1,48 @@
...
+RegisterCommand("switchPlayer", () => {
+    if (IsPlayerSwitchInProgress()) return;
+    const ped = PlayerPedId();
+    SwitchToMultiFirstPart(ped, 0, 1);
+    Delay(5000); // Delay doesn't exist, you have to make it yourself
+    SwitchToMultiSecondPart(ped);
+}, false);
+```
+
+```csharp
```
4mmonium, line 38: `` ```cs `` is more common for docs, but this may work.

4mmonium (changes requested):
Hey, looks good to me for the most part. I requested a couple changes for whenever you're free.

Copy pasted some of my comments for things that are 'redundant'.
Thank you for contributing! 😊

STREAMING/SwitchToMultiFirstpart.md
ahcenezdh: should be good

STREAMING/SwitchToMultiFirstpart.md
ahcenezdh: should be good

STREAMING/SwitchToMultiSecondpart.md
ahcenezdh: should be good

STREAMING/GetPlayerSwitchState.md
ahcenezdh: should be good

4mmonium (approved):
Approved, thank you!

PR #936, opened 2023-11-25
Update GetInteriorHeading.md

freedy69:
Added information about the return value.

### DIFF
--- a/INTERIOR/GetInteriorHeading.md
+++ b/INTERIOR/GetInteriorHeading.md
@@ -1,18 +1,15 @@
 ---
 ns: INTERIOR
 ---
-## _GET_INTERIOR_HEADING
+## GET_INTERIOR_HEADING
 
 ```c
 // 0xF49B58631D9E22D9
-float _GET_INTERIOR_HEADING(int interior);
-```
-
-```
-GET_INTERIOR_*
-
-NativeDB Introduced: v1493
+float GET_INTERIOR_HEADING(int interior);
 ```
 
 ## Parameters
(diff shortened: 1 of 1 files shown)


INTERIOR/GetInteriorHeading.md
```diff
@@ -1,18 +1,15 @@
...
-NativeDB Introduced: v1493
+float GET_INTERIOR_HEADING(int interior);
 ```
 
 ## Parameters
-* **interior**:
+* **interior**: ID of the interior
+
+## Return value
+Returns interior heading in radians. Multiply the returned value with 57.29578 to get it in degrees.
```
4mmonium, line 15: Just a quick note, that `57.29578` is the result of [180/Pi](https://en.wikipedia.org/wiki/Pi) or [360/τ](https://en.wikipedia.org/wiki/Turn_(angle)#Tau_proposals) for people that prefer such.

### Pi = π (3.14159...)
### Tau = 2π (6.28318...)

4mmonium (changes requested):
Hey almost there, you forgot to add the alias for the native since you changed the function name 😊

INTERIOR/GetInteriorHeading.md
freedy69: I changed the return value text, how is it now?

4mmonium: > I changed the return value text, how is it now?

LGTM now, thank you 👍

4mmonium (approved):

PR #939, opened 2023-12-04
Update NetworkResurrectLocalPlayer.md

Ra1n6666:
Found that if I put unk to false it just resurrects me normally but if I put it true it resurrects me to another ped which makes my old ped stay.

Screenshot_1


ahcenezdh:
At this point you can give examples to use the natives (lua, js, c#) and complete arguments field 

```
**x**:
 **y**: 
* **z**: 
* **heading**: 
* **keepBody**: 
* **changetime**: 
```

4mmonium (changes requested):
Some changes needed check what I commented: 

Regarding this:
NativeDB Added Parameter 7: Any p6

I searched native references and it may very well be a boolean, I would name it `bToggleRenderPhases`.

Remember that adding arguments is not supported at the moment, so you can only add a note in the description for newly added arguments 🙂

4mmonium (approved):
Approved after some improvements, thank you for your contribution.

PR #940, opened 2023-12-05
tweak(natives/vehicle): move return value for GetVehicleMod

AvarianKnight:
the parameters [don't currently show in the docs](https://docs.fivem.net/natives/?_0x772960298DA26FDB) this will hopefully fix that


4mmonium (approved):
LGTM, thank you!

PR #947, opened 2023-12-18
feat(natives/vehicles): add new examples and update natives for vehicles

ahcenezdh:
Update various names/descriptions for vehicle natives, and add some examples


ahcenezdh:
Could explain in which case i'm breaking the ABI compatibility please?

Does switching from Any to Vehicle going to break it?

Or switching from void to bool?

4mmonium:
> Could you explain in which case i'm breaking the ABI compatibility please?
> 
> Does switching from Any to Vehicle going to break it?
> 
> Or switching from void to bool?

Switching from `Any` to `int` doesn't break it, but switching from `Any` to any other commonly used type does. The reason behind it is due to signature changes between parameters, all the parameter types together produce a given signature, not using `cs_type` causes the new signature to not be backwards compatible (mainly happens in C# when a user tries to call a native from the API class and the signature to look up is non-existent).

I believe `Any` to `Vehicle` or any other entity type such as `Ped` shouldn't break for that matter, so no it shouldn't be required to use `cs_type` for those kind of changes (or at least I didn't find any in the entire natives repo). I believe it will only break if the change involves commonly used types such as `Any` or `Vehicle` to `float`, `short`, etc (would also need `cs_type` for other cases such as `Vehicle` to `bool` i.e. [this](https://github.com/citizenfx/natives/blob/4b5b140ec3302eda3861edcf3ec77f3be92d47fb/VEHICLE/SetBoatSinksWhenWrecked.md?plain=1#L9)).

Here's [codegen](https://github.com/citizenfx/fivem/blob/bb270ff5cb320831d77a2ca4541159a1dc582362/ext/natives/codegen_out_cs.lua#L156) if you want to look further into it.

ahcenezdh:
Oh okay thanks, so I guess my commits doesn't break ABI compatibility, I'm waiting your review

4mmonium (changes requested):
Hi, I reviewed this and left you some comments 🙂

I don't mean to set a standard here, but I would like to at least experiment breaking these down into different PRs; each with a maximum of say, five to eight file changes (including file removals/adds). This approach would allow for smoother integration of merge/pull requests.

This is somewhat important, especially when compared to the alternative of consolidating all changes into a single PR, where a mistake could significantly impede progress.

Thank you for your contribution! 😊

ahcenezdh:
Hey no problem @4mmonium i'm going to close this PR and re open to seperate everything based on your review!

PR #952, opened 2023-12-21
feat(natives/vehicle): update vehicle natives - Part 3/4

ahcenezdh:
should be good for review @4mmonium


4mmonium (approved):
LGTM, thanks!

PR #957, opened 2023-12-24
feat(natives/gamebuild): update natives from the 2189 gamebuild

ahcenezdh:
Update natives from the 2189 gamebuild


4mmonium (approved):
Reviewed, looks good to me.
Merry Christmas!

PR #959, opened 2023-12-24
feat(natives/gamebuild): enhance natives compatibility with 2189 build

ahcenezdh:
Again i updated natives from the 2189 build, should be good for review @4mmonium


4mmonium (approved):
LGTM, merging, thank you!

PR #964, opened 2023-12-26
Add proper description for IsEntityOnScreen

WithLithum:
Adds a properly formatted description for `IsEntityOnScreen`.

I do not consider an example would be needed for this native considering how simple is the native is, and the new description means about the same thing as of the old imported description.

Feel free to edit this PR if you think something is wrong.

### DIFF
--- a/ENTITY/IsEntityOnScreen.md
+++ b/ENTITY/IsEntityOnScreen.md
@@ -12,5 +12,9 @@ Determines whether the screen position of the specified entity is within the 2D 
 
 This native will not check if the entity is not visible due to being occluded (for example, behind a wall). To check if a entity is on screen and is not occluded, use [IS_ENTITY_OCCLUDED](#_0xE31C2C72B8692B64).
 
+## Return value
+
+Returns `true` if the the entity is in between the minimum and maximum values for the 2d screen coord; otherwise, `false`.
+
 ## Parameters
 * **entity**: The entity to check.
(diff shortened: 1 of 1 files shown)


ahcenezdh:
You need to add the ## Return value and specify what information is returning this native

ENTITY/IsEntityOnScreen.md
```diff
@@ -8,13 +8,9 @@ ns: ENTITY
 BOOL IS_ENTITY_ON_SCREEN(Entity entity);
 ```
 
-```
-Returns true if the entity is in between the minimum and maximum values for the 2d screen coords.
```
PsychoShock, line 12: This should be on the return value instead of deleting.

ENTITY/IsEntityOnScreen.md
```diff
@@ -8,13 +8,13 @@ ns: ENTITY
...
-```
+Determines whether the screen position of the specified entity is within the 2D bounds of the screen.
 
-## Parameters
-* **entity**: 
+This native will not check if the entity is not visible due to being occluded (for example, behind a wall). To check if a entity is on screen and is not occluded, use [IS_ENTITY_OCCLUDED](#_0xE31C2C72B8692B64).
 
 ## Return value
+
+Returns `true` if the the entity is in between the minimum and maximum values for the 2d screen coord; otherwise, `false`.
```
PsychoShock, line 17: Return value need to be after param. Please refer to https://github.com/citizenfx/natives/blob/master/.github/CONTRIBUTING.md

PsychoShock (changes requested):

PsychoShock (approved):
Look's good. Please squash it [like this](https://docs.fivem.net/docs/contributing/git/squash-guide/) to make it cleaner..

PR #966, opened 2023-12-27
Update IsPedDeadOrDying.md

ahcenezdh:
* Update IsPedDeadOrDying documentation with the second arg and explain why it does not "detect" when the ped is in a dying phase

[source](https://vhub.wiki/natives/0x3317DEDB88C95038)


4mmonium (approved):
Approved, thanks!

PR #970, opened 2024-01-01
Rename SetVehicleJetEngineOn -> SetVehicleKeepEngineOnWhenAbandoned

Scaarus:
This is a simple rename with an added description. Calling this native on any vehicle with toggle=true causes the engine of the vehicle to remain on even after exiting. Setting to false simply returns the vehicle to normal behavior.

Found this in the source code that's been floating around this past week and tested to ensure it works as expected.


4mmonium (approved):
LGTM, made some slight nitpicks, that's all.

PR #973, opened 2024-01-06
Update StopGameplayCamShaking.md

TheCoolestPaul:
Tested in a server and found the boolean parameter would stop shaking the frame it was called on, instead of slowing the shake to a stop.

### DIFF
--- a/CAM/StopGameplayCamShaking.md
+++ b/CAM/StopGameplayCamShaking.md
@@ -5,10 +5,9 @@ ns: CAM
 
 ```c
 // 0x0EF93E9F3D08C178 0xFD569E4E
-void STOP_GAMEPLAY_CAM_SHAKING(BOOL bstopImmediately);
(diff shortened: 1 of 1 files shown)


CAM/StopGameplayCamShaking.md
```diff
@@ -5,10 +5,10 @@ ns: CAM
...
 ```c
 // 0x0EF93E9F3D08C178 0xFD569E4E
-void STOP_GAMEPLAY_CAM_SHAKING(BOOL p0);
+void STOP_GAMEPLAY_CAM_SHAKING(BOOL stopImmediately);
 ```
 
 
 ## Parameters
-* **p0**: 
+* **stopImmediately**: Should the shake stop this frame?
```
4mmonium, line 13: It stops immediately, yes (you can remove the question mark 😅). It's preferable to name this `bStopImmediately` since it's a `BOOL` parameter.

4mmonium (changes requested):
Hey there, requested some changes, thank you for your contribution!

PR #976, opened 2024-01-13
fix(weapon): correct melee weapon group hash

someone005:


4mmonium:
Hi, it looks like this was updated in a more recent PR, but thank you for the efforts! 🙂

PR #977, opened 2024-01-18
Better documentation for SetVehicleNumberPlateTextIndex

ItsJunction:
Updates the native to have 'modern docs' along with adding images, indexes and names for plates


PsychoShock (changes requested):
This is great work. I think it would be nicer to have smaller image, you can do it with this code: ``

Also, maybe remove `Indexes as of b3095:` and add a column for the required build. Would be easier for the next updates.

PsychoShock:
Following `imgur`, this PR will have to wait until a solution is given for what to do.

AvarianKnight:
These should be added to the FiveM docs under the vehicle section.

PR #990, opened 2024-02-15
fix: few cutscene links

gottfriedleibniz:
Related to: https://github.com/citizenfx/fivem/pull/2378. 

The hash to `0x011883F41211432A` was also lowercase while all other natives use uppercase.


gottfriedleibniz:
There seems to be a conflict with  https://github.com/citizenfx/natives/blob/master/CUTSCENE/GetCutsceneEndTime.md https://github.com/citizenfx/natives/blob/master/CUTSCENE/N_0x011883f41211432a.md. 

I'll change the hash back to lowercase and that issue can be rectified later.

4mmonium (approved):
LGTM, thank you!

PR #992, opened 2024-02-15
Update SetPedResetFlag to include full enum

JohnDeved:
pr adds full enum and comments for SetPedResetFlag


ahcenezdh:
Would it be better to centralize enums in the fivem-docs repo, @4mmonium? What do you think about that?

4mmonium:
> Would it be better to centralize enums in the fivem-docs repo, @4mmonium? What do you think about that?

I think it's fine to have it in the natives, as long as they're referenced by other files and not copy/pasted all over. However, things such as images (license plates for example among other things) could perhaps fit in the game references section of the docs.

Issue #989, opened 2024-02-13
Update N_0x83a169eabcdb10a2


CritteRo:
This native is referenced as `PED::SET_PED_PHONE_PALETTE_IDX` in nativedb: https://alloc8or.re/gta5/nativedb/?n=0x83A169EABCDB10A2

Would be nice to have the native name changed, and also documented, so future users won't need to dig through old code to find the native.

It has 2 params, first param is the ped ID, second is the ID of the phone palette. The pallets mimic the IDs of CELLPHONE_IFRUIT themes: 
```
0=blue,
1=green,
2=red,
3=orange,
4=gray,
5=purple,
6=pink,
7 or more = dark-gray?
```

Current lua example:
```lua
N_0x83a169eabcdb10a2(PlayerPedId(), 3)
```

Example image of the phone case being changed:
![image](https://github.com/citizenfx/natives/assets/64140178/09c7ce9f-3f44-49f6-809c-3072939b6673)

PsychoShock:
You are free to PR the modification. You can check [this](https://github.com/citizenfx/natives/blob/master/.github/CONTRIBUTING.md) if you never PR a native! If you do something wrong, you'll be told what's the issue and how to fix it :)

Contributions are always welcome in FiveM / RedM.

ahcenezdh:
> This native is referenced as `PED::SET_PED_PHONE_PALETTE_IDX` in nativedb: https://alloc8or.re/gta5/nativedb/?n=0x83A169EABCDB10A2
> 
> Would be nice to have the native name changed, and also documented, so future users won't need to dig through old code to find the native.
> 
> It has 2 params, first param is the ped ID, second is the ID of the phone palette. The pallets mimic the IDs of CELLPHONE_IFRUIT themes:
> 
> ```
> 0=blue,
> 1=green,
> 2=red,
> 3=orange,
> 4=gray,
> 5=purple,
> 6=pink,
> 7 or more = dark-gray?
> ```
> 
> Current lua example:
> 
> ```lua
> N_0x83a169eabcdb10a2(PlayerPedId(), 3)
> ```
> 
[...]

PR #993, opened 2024-02-15
feat(natives/ped): add SetPedPhonePaletteIdx

ahcenezdh:
For js & c# examples i'm not sure at all, someone have to review it to be sure. Close the issue #989


4mmonium (changes requested):
Hey there, this replaces a native, therefor the old [native file](https://github.com/citizenfx/natives/blob/f1f7a7d3f438da81d006bf0dfb59012d40c8272e/PED/N_0x83a169eabcdb10a2.md?plain=1#L4) should be removed. Ideally, an alias pointing to the old native name should be added as well since that's standard.

PED/SetPedPhonePaletteIdx.md
```diff
@@ -0,0 +1,127 @@
...
+```c
+// 0x83A169EABCDB10A2
+void SET_PED_PHONE_PALETTE_IDX(Ped ped, int index);
+```
+
+Sets the palette index of a ped's phone.
+
+| Value | Color   |
+| :---: | :-----: |
+|  `0`  | Blue    |
```
PsychoShock, line 15: Actually `0` is light Blue theme from the settings menu.

ahcenezdh:
> Hey there, this replaces a native, therefor the old [native file](https://github.com/citizenfx/natives/blob/f1f7a7d3f438da81d006bf0dfb59012d40c8272e/PED/N_0x83a169eabcdb10a2.md?plain=1#L4) should be removed. Ideally, an alias pointing to the old native name should be added as well since that's standard.

For x reason when searching it on my visual studio code, couldn't find it

PED/SetPedPhonePaletteIdx.md
ahcenezdh: 👍

ahcenezdh:
Should be good

ahcenezdh:
Hey @4mmonium srry to ping you but any news for this PR?

PED/SetPedPhonePaletteIdx.md
```diff
@@ -0,0 +1,128 @@
...
+        print("Color not found: " .. color) -- Color doesn't exist in the table. 
+        return -- Exits the function.
+    end
+
+    -- Use the native to set the ped's phone palette index based on the selected color.
+    SetPedPhonePaletteIdx(playerPed, colors[color])
+end
+
+-- Call the function with "Green" to set the phone color to green for the player
+setPhoneColor("Green")
```
4mmonium, line 61: This is neat, but it just complicates the example for the average newcomer. Don't introduce logic that could result complex for newcomers, simply define the `enum` for the colors somewhere and refer to the `enum`. Same applies for the other examples. 

This is great, but why doing this when the native takes an integer anyway? Both would achieve the same result, just less complex and more straight to the point.

Thanks!

4mmonium (changes requested):
This is OK, but I'd prefer if you used an `enum` rather than passing it as a `string` for hash table (key-value pair tables) lookup. Check my comment, thanks!

4mmonium (approved):
Looks good to me, merging, thank you!

PR #1000, opened 2024-02-16
feat(audio): batch 4 of native updates

AvarianKnight:


AUDIO/GetNextAudibleBeat.md
```diff
@@ -0,0 +1,19 @@
...
+aliases: ["0xC64A06D939F826F5"]
+---
+## GET_NEXT_AUDIBLE_BEAT
+
+```c
+// 0xC64A06D939F826F5
+BOOL GET_NEXT_AUDIBLE_BEAT(float* timeInSeconds, cs_type(AnyPtr) float* bpm, int* beatNum);
+```
+
+timeS:
```
PsychoShock, line 12: Typo on `S`

AUDIO/GetSoundId.md
```diff
@@ -8,5 +8,18 @@ ns: AUDIO
 int GET_SOUND_ID();
 ```
 
+This native returns a new SoundId, which is used for keeping track of sounds after
+they've been triggered use this if you need to control a sound after it's been started,
+for instance to stop a looping sound, or to change a sound's pitch midway through playback.
```
PsychoShock, line 13: This should be on the return section and not description.

AUDIO/GetStreamPlayTime.md
```diff
@@ -8,5 +8,5 @@ ns: AUDIO
 int GET_STREAM_PLAY_TIME();
 ```
 
-
 ## Return value
+Returns the stream play time
```
PsychoShock, line 12: `Return the current streamed play time` would be a better return description.

4mmonium (approved):
LGTM, thanks!

PR #1002, opened 2024-02-16
feat(audio): batch 6 of native updates

AvarianKnight:


AUDIO/IsAlarmPlaying.md
```diff
@@ -8,12 +8,8 @@ ns: AUDIO
 BOOL IS_ALARM_PLAYING(char* alarmName);
 ```
 
-```
-Example:  
-bool playing = AUDIO::IS_ALARM_PLAYING("PORT_OF_LS_HEIST_FORT_ZANCUDO_ALARMS");  
-```
```
PsychoShock, line 14: Instead of removing it, you can maybe add an example on the good section? :)

AvarianKnight: These batches of PR are just meant to sync the natives with the docs, adding example sections and such can be done in future PRs.

PR #1027, opened 2024-02-27
Document NetworkAllowLocalEntityAttachment

freedy69:
Changed to real name and added documentation


4mmonium (approved):
Nice find, the name for this was incorrect, thanks for the fix!

PR #1043, opened 2024-03-07
Update map natives

freedy69:


HUD/IsPausemapInInteriorMode.md
```diff
@@ -0,0 +1,13 @@
...
+---
+## IS_PAUSEMAP_IN_INTERIOR_MODE
+
+```c
+// 0x9049FE339D5F6F6F 0x199DED14
+BOOL IS_PAUSEMAP_IN_INTERIOR_MODE();
+```
+
+## Return value
+Returns if the map is in an interior.
```
PsychoShock, line 13: No need to have a return value description when native name is obvious.

HUD/PauseToggleFullscreenMap.md
```diff
@@ -0,0 +1,16 @@
...
+aliases: ["0x2DE6C5E2E996F178"]
+---
+## PAUSE_TOGGLE_FULLSCREEN_MAP
+
+```c
+// 0x2DE6C5E2E996F178
+void PAUSE_TOGGLE_FULLSCREEN_MAP(cs_type(Any) BOOL enabled);
+```
+
+Toggles pause menu map rendering. Please note that this native is EXTREMELY buggy so you have to be very careful while using it, since it can easily cause crashes if you're using this while the menu content is being navigated.
```
PsychoShock, line 12: Could you provide a crash dump instead? Maybe this is a issue that can be fix.

HUD/SetFakePausemapPlayerPositionThisFrame.md
```diff
@@ -0,0 +1,32 @@
...
+aliases: ["0x77E2DD177910E1CF", "_SET_PLAYER_BLIP_POSITION_THIS_FRAME"]
+---
+## SET_FAKE_PAUSEMAP_PLAYER_POSITION_THIS_FRAME
+
+```c
+// 0x77E2DD177910E1CF 0x54E75C7D
+void SET_FAKE_PAUSEMAP_PLAYER_POSITION_THIS_FRAME(float x, float y);
+```
+
+Overrides the position of the main player blip for one frame.
```
PsychoShock, line 12: for the current frame*

PsychoShock (commented):
Hey,

I have left some comments if you wanna check it out. Nice work.

HUD/SetFakePausemapPlayerPositionThisFrame.md
freedy69: does such a small difference really matter?

HUD/IsPausemapInInteriorMode.md
freedy69: okay, thank you!

PsychoShock:
Since you removed
> Please note that this native is EXTREMELY buggy so you have to be very careful while using it, since it can easily cause crashes if you're using this while the menu content is being navigated.

Can you try to have a repo so I can try to get crash too?

freedy69:
> Since you removed
> 
> > Please note that this native is EXTREMELY buggy so you have to be very careful while using it, since it can easily cause crashes if you're using this while the menu content is being navigated.
> 
> Can you try to have a repo so I can try to get crash too?

as i remember all i did was call the native when the map was fullscreen, maybe i was in the interior though i tested both cases and they were fine now :shrug:

4mmonium (approved):
LGTM, remember to use threads when calling non-stop while loops, newcomers might not be aware of this.

Thank you for your contribution! 😊

PR #1051, opened 2024-03-09
chore(natives): standardize threading with Citizen

ahcenezdh:
Standardized threading and waiting functions for consistency across the documentation.

### DIFF
--- a/CAM/SetCamDofStrength.md
+++ b/CAM/SetCamDofStrength.md
@@ -16,7 +16,7 @@ Specifies how much the DoF effect should be applied (Set using [`SET_CAM_NEAR_DO
 
 ## Examples
 ```lua
-CreateThread(function()
+Citizen.CreateThread(function()
     local camera = CreateCam("DEFAULT_SCRIPTED_FLY_CAMERA", true)
     -- Set the cam coordinates to the player coords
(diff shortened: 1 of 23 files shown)


AvarianKnight:
Would it not be better to prefer the short forms over ones with Citizen prefix?

ahcenezdh:
> Would it not be better to prefer the short forms over ones with Citizen prefix?

Good question! Honestly, I did it with Citizen, since in most examples, there is Citizen, and a lot of people are using it.

PsychoShock:
Hello,

There's no point of doing this since it's all the same. You can check https://github.com/citizenfx/fivem/blob/master/data/shared/citizen/scripting/lua/scheduler.lua#L136. Also, it will be most common to remove it that adding it back. Only reason, some has `Citizen` is from old PR's.

Thanks.

ahcenezdh:
Hello, I'm aware of that. The objective, as indicated by the title of this PR, is to standardize the use of "Citizen" (though it could also be without "Citizen") for greater consistency. This approach aims to prevent new scripters on FiveM from getting lost. It would be better for the documentation if we choose to stick with only one of the two options either with "Citizen" or without.

PR #1057, opened 2024-03-12
feat(audio/IsVehicleRadioOn): update examples

ahcenezdh:
* Update IS_VEHICLE_RADIO_ON examples


4mmonium (commented):
Hi, I don't think I prefer this over the previous PR, the previous `C#` code provides for more use-cases. Ideally the developer might want to get the vehicle for any other player and not just the local one.

I think the old PR is fine, open to hearing your feedback though.

Thanks!

ahcenezdh:
Hi no problem, sounds good for me!

PR #1058, opened 2024-03-12
feat(vehicle/SetVehicleDoorsLocked): update c# example

ahcenezdh:
Update SET_VEHICLE_DOORS_LOCKED examples on the C# version to use the c# wrapper, thanks to Quadruplex & VIRUXE for explaining me how is it working

Someone should update the enum in the [c# wrapper](https://github.com/citizenfx/fivem/blob/8b93ef263f644196fd83621df65c3c0b687da124/code/client/clrcore/External/Vehicle.cs#L307) btw


ahcenezdh:
Maybe we should the add the version without the c# wrapper?

PR #1062, opened 2024-03-16
feat(natives/vdist): update vdist/vdist2 natives

ahcenezdh:
* Update VDIST
* Update VDIST2


freedy69:
Great PR, maybe if possible you could add examples of the faster mathematical calculation functions as well?

ahcenezdh:
> Great PR, maybe if possible you could add examples of the faster mathematical calculation functions as well?

Thanks you, indeed could be better to add examples for this, i'm going to do it

4mmonium (approved):
LGTM, merging, thank you!

PR #1071, opened 2024-03-20
feat(natives/PED): Add naming and documentation for unnamed natives

theskiratta:
Using the amazing Alloc8or's [native database](https://alloc8or.re/gta5/nativedb/), I added naming and some documentation for **nearly all** previously unnamed ped natives. The only one I skipped is 0x83a169eabcdb10a2 / SetPhonePaletteIDX, as I noticed this was already worked on as stated [here](https://github.com/citizenfx/natives/issues/989). 

I tried my best to do this as is explained in the guidelines and the examples given within those. I'm happy to fix any mistakes I may have made.

Example of change:
0x5407B7288D0478B7 was updated to it's proper name, COUNT_PEDS_IN_COMBAT_WITH_TARGET. I explained what the native does, what should be passed as a parameter and what should be expected as a return. Further, I linked to a similar native (COUNT_PEDS_IN_COMBAT_WITH_TARGET_WITHIN_RADIUS).

I tried to keep one commit per native to make changes and review easier.


PsychoShock (changes requested):
Hello,

This is a big work, but there's too much review to do in a single PR and can mostly broke some nataives. Problem is, no revert would be possible. As @4mmonium said in the past, the prefered maximum is 7-8 when new natives are added. It's easier to review them / merge them quickly. Feel free to do multiples PR's and tag me, I don't mind checking them to help. Also a couple thing to be carreful:

- When a native is not used, you can say it's "decapred" on descripton
- External links are not recommanded / Feel free to PR things in FiveM game docs if needed
-  Some natives seems to have the wrong params. This is the problem when copy / paste natives from external websites

Just saying that this is a big work and it's always appreciate to see other members offering free time. Please note that this is my personnal opinion and I might be wrong here.

Thanks

theskiratta:
Thanks for the comment @PsychoShock ,

I wasn't aware of the 7-8 natives per PR, though I can understand why it would be wanted. I'm not an expert with GitHub, how would I go about making a PR that doesn't include all the commits I've made inside my fork, but only some, so that I can follow this guidance?
Separate to that, please let me know what errors you noticed, specifically with the wrong params, so I may fix them?

Best Regards

Mathu-lmn:
Hey, you could create a new branch of the fork for each PR, and use the [cherry-pick command](https://git-scm.com/docs/git-cherry-pick)

AvarianKnight:
Thank you for the PR.

We're currently not accepting batch updates to natives.

PR #1079, opened 2024-04-03
Use real name for GET_OBJECT_OFFSET_FROM_COORDS

freedy69:
Also added description to offset parameters


ahcenezdh (commented):
For the description of the native, this should be better
> Calculates the world coordinates after applying the specified offsets to the given position, relative to a certain heading. This native is similar to `GET_OFFSET_FROM_OBJECT_IN_WORLD_COORDS` but uses a world position and heading as the reference point.

You can put a description for vector parameters
> xPos: The X coordinate of the original position.
yPos: The Y coordinate of the original position.
zPos: The Z coordinate of the original position.
heading: The heading (in degrees) relative to which the offsets are applied.

For the return value you can say this
> Returns a Vector3 representing the new world coordinates after applying the offsets to the original position relative to the specified heading.

freedy69:
> For the description of the native, this should be better
> 
> > Calculates the world coordinates after applying the specified offsets to the given position, relative to a certain heading. This native is similar to `GET_OFFSET_FROM_OBJECT_IN_WORLD_COORDS` but uses a world position and heading as the reference point.
> 
> You can put a description for vector parameters
> 
> > xPos: The X coordinate of the original position.
> > yPos: The Y coordinate of the original position.
> > zPos: The Z coordinate of the original position.
> > heading: The heading (in degrees) relative to which the offsets are applied.
> 
> For the return value you can say this
> 
> > Returns a Vector3 representing the new world coordinates after applying the offsets to the original position relative to the specified heading.

thank ou so much, done

4mmonium (approved):
LGTM, merging, thank you!

PR #1088, opened 2024-04-21
Update SetPedCanPlayAmbientIdles.md

freedy69:
Added parameter info and also description

### DIFF
--- a/TASK/SetPedCanPlayAmbientIdles.md
+++ b/TASK/SetPedCanPlayAmbientIdles.md
@@ -6,7 +6,7 @@ aliases: ["0x8FD89A6240813FD0"]
 
 ```c
 // 0x8FD89A6240813FD0 0xB5AD044E
-void SET_PED_CAN_PLAY_AMBIENT_IDLES(Ped ped, BOOL disableIdleAnims, BOOL stopCurrentAnim);
+void SET_PED_CAN_PLAY_AMBIENT_IDLES(Ped ped, BOOL bDisableIdleAnims, BOOL bStopCurrentAnim);
 ```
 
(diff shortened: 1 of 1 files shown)


PsychoShock (commented):
Hello,

For better docs, can you change the params by adding `b` before to understand it's a bool.

Thanks :)

TASK/SetPedCanPlayAmbientIdles.md
```diff
@@ -6,18 +6,14 @@ aliases: ["0x8FD89A6240813FD0"]
...
 ```
 
-```
-Appears only in fm_mission_controller and used only 3 times.  
-ped was always PLAYER_PED_ID()  
-p1 was always true  
-p2 was always true  
-```
+Prevents a ped from playing ambient idle animations.
+This native must be called every frame.
```
PsychoShock, line 13: Can you change the description in this format:

```
Prevents a ped from playing ambient idle animations.

Note: This native must be called every frame.
```

freedy69:
> Hello,
> 
> 
> 
> For better docs, can you change the params by adding `b` before to understand it's a bool.
> 
> 
> 
> Thanks :)

Sure but im confused because some commits have those and some dont?

TASK/SetPedCanPlayAmbientIdles.md
freedy69: To a codeblock?

PsychoShock:
Yeah well it's suposse to be a "standart" to have every native with this following but it's hard to keep track on everyone of them. This is just my personnal opinion, maybe @4mmonium don't mind.

4mmonium (approved):
Thank you for your contribution, approved!

PR #1091, opened 2024-04-22
Document various network synchronised scene natives

freedy69:
Added return value, description and parameter information. Also renamed some parameters to what they actually do.


4mmonium (approved):
LGTM, thank you. Merging!

PR #1095, opened 2024-04-25
fix(doc/blip): correct link and param

PsychoShock:
Reference for the link should be [this](https://docs.fivem.net/natives/?_0xB98236CAAECEF897) instead of the current link commit.

This reverts citizenfx/natives#1060 commit

### DIFF
--- a/HUD/RemoveBlip.md
+++ b/HUD/RemoveBlip.md
@@ -5,7 +5,7 @@ ns: HUD
 
 ```c
 // 0x86A652570E5F25DD 0xD8C3C1CD
-void REMOVE_BLIP(int* blip);
+void REMOVE_BLIP(int blip);
 ```
 
 Removes the blip from your map.
--- a/HUD/RemoveBlip.md
+++ b/HUD/RemoveBlip.md
@@ -10,8 +10,6 @@ void REMOVE_BLIP(Blip* blip);
 
 Removes the blip from your map.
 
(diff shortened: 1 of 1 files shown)


AvarianKnight:
Blip isn't a char* though, its an int.

PsychoShock:
Yep thanks, no idea what I was thinking with `char`.

freedy69:
why not keep it as Blip since all blip natives use that type?

AvarianKnight:
As freddy said, changing the type here is un-needed.

HUD/RemoveBlip.md
```diff
@@ -5,12 +5,12 @@ ns: HUD
 
 ```c
 // 0x86A652570E5F25DD 0xD8C3C1CD
-void REMOVE_BLIP(Blip* blip);
+void REMOVE_BLIP(int blip);
```
AvarianKnight, line 8:
```suggestion
void REMOVE_BLIP(Blip* blip);
```

AvarianKnight (changes requested):

HUD/RemoveBlip.md
PsychoShock: No it's required, like you can see here docs doesn't handle correctly the params:
image

This will fix the visual:
image


Old issue, **this was requested by the team before making the PR**.

AvarianKnight: This doesn't "fix" it, this works around it, and subsequently breaks ABI compatibility with C# **which you cannot do**.

AvarianKnight:
This change would break compatibility with the C# runtime by changing the C# codegen from `RemoveBlip(ref blipIndex)` -> `RemoveBlip(blipIndex)`

PR #1096, opened 2024-04-25
fix(audio): fix enum not having a closing tag.

AvarianKnight:
Noticed this when scrolling through natives, fixes my oversight in one of the audio batch updates


4mmonium:
Whoops, nice catch, totally missed it, merging!

PR #1097, opened 2024-04-25
tweak(natives/audio): move comment to line above

AvarianKnight:
This is mainly a stylistic change to get rid of the side-scroll on the native


alberttheprince:
It would be worth adding information about how this only works for non-player peds. Spent an unreasonable amount of time trying to get this to work with player peds only to find that it doesn't work with player peds, regardless of setting voice groups etc.

PR #1099, opened 2024-04-25
Use real names for remaining scaleform natives

freedy69:


GRAPHICS/GetScaleformMovieMethodReturnValueBool.md
```diff
@@ -1,19 +1,18 @@
...
 ```c
 // 0xD80A80346A45D761
-BOOL _GET_SCALEFORM_MOVIE_METHOD_RETURN_VALUE_BOOL(int methodReturn);
+BOOL GET_SCALEFORM_MOVIE_METHOD_RETURN_VALUE_BOOL(int methodReturn);
 ```
 
-```
-methodReturn: The return value of this native: END_SCALEFORM_MOVIE_METHOD_RETURN_VALUE
-```
+Returns a boolean value that is returned by a scaleform movie's actionscript.
```
AvarianKnight, line 12:
```suggestion

```
This mimics `## Return value`, which is un-needed

GRAPHICS/GetScaleformMovieMethodReturnValueBool.md
```diff
@@ -1,19 +1,18 @@
...
 
 ## Parameters
-* **methodReturn**: 
+* **methodReturn**: Operation handle returned by [`END_SCALEFORM_MOVIE_METHOD_RETURN_VALUE`](#_0xC50AA39A577AF886)
 
 ## Return value
+The return value passed from actionscript to game code.
```
AvarianKnight, line 18: This is very confusing, but I also can't think of a better way to word this, if you can please edit it, otherwise just mark this as resolved.

AvarianKnight (changes requested):

GRAPHICS/GetScaleformMovieMethodReturnValueBool.md
freedy69: i'll keep it for now, couldnt think of anything else to put there sadly

PR #1101, opened 2024-04-26
Real name for 0xEA2F2061875EED90

freedy69:
Used real name and return value information to avoid confusion since the old name is wrong


4mmonium (approved):
Approved, merging, thank you for your contribution!

PR #1106, opened 2024-05-07
feat(natives/graphics): update screens getters docs

ahcenezdh:
Should be good


4mmonium (approved):
LGTM, made slight corrections, thank you!

PR #1110, opened 2024-05-07
Rename .md files according to the native name

Mathu-lmn:
Some files have the wrong name compared to the native's name, this PR fixes it


4mmonium (approved):
Merging, thank you!

PR #1113, opened 2024-05-12
Update PauseMenuSetBusySpinner

freedy69:
Added parameters and description.


MonsieurBibo:
I believe you need to use `cs_type(Any) int` instead of `int` to not break ABI. Though I might be wrong, we should wait the opinion of someone experienced about that.

freedy69:
> I believe you need to use `cs_type(Any) int` instead of `int` to not break ABI. Though I might be wrong, we should wait the opinion of someone experienced about that.

iirc it parses as int or something like that so it's fine for Any to int but other types yea it needs cs_type

4mmonium (approved):
LGTM, thank you for your contribution!

PR #1115, opened 2024-05-12
Update IsVehicleStoppedAtTrafficLights

Mathu-lmn:
This behavior was checked using this code : 
```lua
    while checkveh do
        Citizen.Wait(0)
        for k, v in pairs(GetGamePool("CVehicle")) do
            if DoesEntityExist(v) then
                local coords = GetEntityCoords(v)
                local dst = #(GetEntityCoords(PlayerPedId()) - coords)
                if dst <= 150.0 then
                    if IsEntityOnScreen(v) then
                        local stopped = IsVehicleStoppedAtTrafficLights(v)
                        DrawText3D(coords.x, coords.y, coords.z, "Stopped: " .. tostring(stopped))
                    end
                end
            end
        end
    end
end)
```

### DIFF
--- a/VEHICLE/IsVehicleStoppedAtTrafficLights.md
+++ b/VEHICLE/IsVehicleStoppedAtTrafficLights.md
@@ -8,11 +8,11 @@ ns: VEHICLE
 BOOL IS_VEHICLE_STOPPED_AT_TRAFFIC_LIGHTS(Vehicle vehicle);
 ```
 
-```
-This native checks if the given vehicle is stopped at a red light.
-
-**Note:** Will return false if the vehicle is stopped behind another vehicle waiting at a traffic light.
-```
+This native checks if the given vehicle is stopped at a red or amber traffic light junction, provided the driver's personality is set to not run amber lights.
 
 ## Parameters
-* **vehicle**:
+* **vehicle**: The vehicle handle.
+
+## Return value
+Will return `true` if the vehicle is at an amber light junction and the vehicle driver's personality is set to not run amber lights, or if the vehicle is at a red light, regardless of the driver's personality. 
+Will return `false` if the vehicle is stopped behind another vehicle waiting at a traffic light.
--- a/VEHICLE/IsVehicleStoppedAtTrafficLights.md
+++ b/VEHICLE/IsVehicleStoppedAtTrafficLights.md
@@ -9,10 +9,11 @@ BOOL IS_VEHICLE_STOPPED_AT_TRAFFIC_LIGHTS(Vehicle vehicle);
 ```
 
 ```
-Is this for red lights only?  more testing required.
+This native checks if the given vehicle is stopped at a red light (this won't return true if the vehicle is stopped behind another vehicle waiting at a traffic light)
 ```
 
 ## Parameters
 * **vehicle**: 
 
 ## Return value
+Returns true if the vehicle is stopped because of a redlight, false otherwise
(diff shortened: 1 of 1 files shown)


VEHICLE/IsVehicleStoppedAtTrafficLights.md
```diff
@@ -9,10 +9,11 @@ BOOL IS_VEHICLE_STOPPED_AT_TRAFFIC_LIGHTS(Vehicle vehicle);
...
 ```
-Is this for red lights only?  more testing required.
+This native checks if the given vehicle is stopped at a red light (this won't return true if the vehicle is stopped behind another vehicle waiting at a traffic light)
 ```
 
 ## Parameters
 * **vehicle**: 
 
 ## Return value
+Returns true if the vehicle is stopped because of a redlight, false otherwise
```
PsychoShock, line 19: No need for a return value when native name is clear.

VEHICLE/IsVehicleStoppedAtTrafficLights.md
```diff
@@ -9,10 +9,11 @@ BOOL IS_VEHICLE_STOPPED_AT_TRAFFIC_LIGHTS(Vehicle vehicle);
 ```
 
 ```
-Is this for red lights only?  more testing required.
+This native checks if the given vehicle is stopped at a red light (this won't return true if the vehicle is stopped behind another vehicle waiting at a traffic light)
```
PsychoShock, line 12: Change for:

Check if given vehicle is stopped at a red light. 

**Note:** will return `false` if the vehicle is stopped behind another vehicle waiting at a traffic light.

In my opinion.

PsychoShock (changes requested):

Mathu-lmn:
Thanks for your feedback, does this looks good to you ?

VEHICLE/IsVehicleStoppedAtTrafficLights.md
```diff
@@ -9,10 +9,10 @@ BOOL IS_VEHICLE_STOPPED_AT_TRAFFIC_LIGHTS(Vehicle vehicle);
 ```
 
 ```
-Is this for red lights only?  more testing required.
+This native checks if the given vehicle is stopped at a red light.
+
+**Note:** Will return false if the vehicle is stopped behind another vehicle waiting at a traffic light.
```
PsychoShock, line 14: false -> `false`

Mathu-lmn: This won't work since it's already in a \`\`\` box no ?

4mmonium: Imported comments shouldn't be modified, instead the fenced code block should be removed and documentation should be written instead, as a description of the native. That way everyone knows the documentation has been updated, it will show "Has modern docs" in the native reference site.

VEHICLE/IsVehicleStoppedAtTrafficLights.md
```diff
@@ -9,10 +9,10 @@ BOOL IS_VEHICLE_STOPPED_AT_TRAFFIC_LIGHTS(Vehicle vehicle);
 ```
 
 ```
```
4mmonium, line 11: Remove backticks, these are reserved for imported comments, not actual contributions by the community.

VEHICLE/IsVehicleStoppedAtTrafficLights.md
```diff
@@ -9,10 +9,10 @@ BOOL IS_VEHICLE_STOPPED_AT_TRAFFIC_LIGHTS(Vehicle vehicle);
...
-Is this for red lights only?  more testing required.
+This native checks if the given vehicle is stopped at a red light.
+
+**Note:** Will return false if the vehicle is stopped behind another vehicle waiting at a traffic light.
 ```
```
4mmonium, line 15: Remove these as well.

4mmonium (changes requested):
Requested changes, check my comments. Thank you for your contribution!

4mmonium:
Made some changes based on research, merging. Thanks!

PR #1116, opened 2024-05-12
Update GetBestPedWeapon

Mathu-lmn:
This was the code used to identify the bool use : 
```lua
    while true do
        Wait(1000)
        print("Boolean off : " .. GetBestPedWeapon(PlayerPedId(), false), "Boolean on : " ..GetBestPedWeapon(PlayerPedId(), true))
    end
end)
```

And here is the video of this code that shows the described behavior : 

https://github.com/citizenfx/natives/assets/80094438/927438ee-5d00-4215-a1db-c1e7f22187be

### DIFF
--- a/WEAPON/GetBestPedWeapon.md
+++ b/WEAPON/GetBestPedWeapon.md
@@ -8,14 +8,9 @@ ns: WEAPON
 Hash GET_BEST_PED_WEAPON(Ped ped, BOOL ignoreAmmoCount);
 ```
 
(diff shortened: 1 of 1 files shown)


Mathu-lmn:
Thanks for your feedback, does this looks good to you ?

PsychoShock (approved):

4mmonium (approved):
LGTM, thank you!

PR #1133, opened 2024-06-20
Update NetworkConcealPlayer.md

dotmbf:
Added more defined example and explanation of use in the native.

### DIFF
--- a/NETWORK/NetworkConcealPlayer.md
+++ b/NETWORK/NetworkConcealPlayer.md
@@ -28,28 +28,39 @@ This setup keeps instance players separate from each other while allowing intera
 * **bAllowDamagingWhileConcealed**: 
 
 ```lua
+function GetPlayerInstance(player)
+	-- you can replace this with your own data
+    local playerServerId = GetPlayerServerId(player)
+    return Player(playerServerId).state.instance_id or 0
+end
+
+local playerId = PlayerId()
 -- Function to manage player visibility
-function concealPlayers(instanceId)
-    local allPlayers = GetPlayers()
+function concealPlayers()
+    local allPlayers = GetActivePlayers()
+	local localPlayerInstance = GetPlayerInstance(playerId)
 
     for _, player in ipairs(allPlayers) do
-        local playerInstance = GetPlayerInstance(player)  -- You need to define this function
+        if player == playerId then goto continue end
 
-        if instanceId == nil then
-            -- General population: hide players in any instance
-            if playerInstance ~= nil then
-                NetworkConcealPlayer(player, true, false)
-            else
-                NetworkConcealPlayer(player, false, false)
-            end
-        else
-            -- Instance players: hide players not in the same instance
-            if playerInstance ~= instanceId then
-                NetworkConcealPlayer(player, true, false)
-            else
-                NetworkConcealPlayer(player, false, false)
-            end
-        end
+        local playerInstance = GetPlayerInstance(player) 
+
+		if playerInstance == localPlayerInstance then
+			-- if we're in the same instance then we want to be able to see them
+			NetworkConcealPlayer(player, false, false)
+		else
+			-- they're in a different instance, so hide them
+			NetworkConcealPlayer(player, true, false)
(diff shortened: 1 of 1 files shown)


NETWORK/NetworkConcealPlayer.md
```diff
@@ -15,3 +15,42 @@ This is what R* uses to hide players in MP interiors.
...
+This setup keeps instance players separate from each other while allowing interaction with the general population.
+
+
+```lua
+-- Function to manage player visibility
+function concealPlayers(instanceId)
+    local allPlayers = GetPlayers()
+
+    for _, player in ipairs(allPlayers) do
+        local playerInstance = GetPlayerInstance(player)  -- You need to define this function
```
4mmonium, line 37: You don't need an instance, not sure why this method exists.
All you need is a valid player index and that player needs to be networked, also known as a networked object or `netObj`.

4mmonium (changes requested):
Hey, requested some changes, thanks!

NETWORK/NetworkConcealPlayer.md
dotmbf: Yeah, I'm aware it's not required, I was just showcasing a use case scenario, for interiors (like GTA:O apartments)

dotmbf: It was to provide an efficient scenario, with interior instancing for people wanting to use the general population (which to my knowledge, is not supported by routing bucket?)

Mathu-lmn: Hmm, I still don't think it's a good idea to put undefined functions in the documentation.

dotmbf: That's very true, but it's very hard to provide an example without some kind of pseudo code

AvarianKnight: It might be better to use a state bag in this case, or define the function with a hard coded instance id and let the end user know this is where they should add how they do their instances

Something like this might also work

```lua
function GetPlayerInstance(player)
    local playerServerId = GetPlayerServerId(player)
    return Player(playerServerId).state.instance_id or 0
end
```

You also don't need the extra if/else branch, you can just check if the instance ids match and if they don't just conceal them.

AvarianKnight (changes requested):

NETWORK/NetworkConcealPlayer.md
dotmbf: The else was added in case that people left the "instance" they would have to be removed from the NET_CONCEAL filter in R*

State bags are also correct, yes, I just wrote what I assumed to be the best example for a simple use-case scenario

NETWORK/NetworkConcealPlayer.md
```diff
@@ -5,13 +5,51 @@ ns: NETWORK
...
+        else
+            -- Instance players: hide players not in the same instance
+            if playerInstance ~= instanceId then
+                NetworkConcealPlayer(player, true, false)
+            else
+                NetworkConcealPlayer(player, false, false)
+            end
+        end
+    end
+end
```
AvarianKnight, line 54:
```suggestion
function GetPlayerInstance(player)
	-- you can replace this with your own data
    local playerServerId = GetPlayerServerId(player)
    return Player(playerServerId).state.instance_id or 0
end

local playerId = PlayerId()
-- Function to manage player visibility
function concealPlayers()
    local allPlayers = GetActivePlayers()
	local localPlayerInstance = GetPlayerInstance(playerId)

    for _, player in ipairs(allPlayers) do
        if player == playerId then goto continue end

        local playerInstance = GetPlayerInstance(player)  -- You need to define this function

		if playerInstance == localPlayerInstance then
			-- if we're in the same instance then we want to be able to see them
			NetworkConcealPlayer(player, false, false)
		else
			NetworkConcealPlayer(player, true, false)
		end

		::continue::
    end
end

CreateThread(function()
	while true do
		concealPlayers()
		Wait(250)
	end
end)
```

PR #1136, opened 2024-06-24
Update IsVehicleTyreBurst.md (indexes for 8 wheels)

iSentrie:
Added indexes for vehicles with 8 wheels.

image

image

### DIFF
--- a/VEHICLE/IsVehicleTyreBurst.md
+++ b/VEHICLE/IsVehicleTyreBurst.md
@@ -18,7 +18,9 @@ completely - is to check if tire completely gone from rim.
 '4 = wheel_lr / bike rear / in 6 wheels trailer, plane or jet is last one on left  
 '5 = wheel_rr / in 6 wheels trailer, plane or jet is last one on right  
 '45 = 6 wheels trailer mid wheel left  
-'47 = 6 wheels trailer mid wheel right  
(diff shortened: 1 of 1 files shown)


colistro123 (changes requested):
Looks invalid as is to me. Provided feedback.

iSentrie:
Looking deeper in to this, shows that each game build has different ID's for 2 wheels.

AvarianKnight:
> Looking deeper in to this, research shows that each game build has different ID's for 2 wheels.
> 
> For example: Build `2802` = 547 / 549 Build `2944` = 412 / 416 Build `3095` = 788 / 792 and so on...

Given the fact that these change per game build, I would say that it wasn't intended to work in the first place.

It is probably better to have this as the default enum than trying to update to a magic value every game update, especially considering there are *very very* few vehicles that will ever need this (iirc it is *only* this vehicle)

From testing all of these values work on b3258 for the 3rd row of tires:

687	1
1039	1
2170	1
3139	1
3357	1
4630	1

All of these work for `0`
```
0	1
6	1
454	1
510	1
594	1
675	1
1027	1
2146	1
3127	1
3203	1
3351	1
3451	1
3719	1
4040	1
4614	1
```

There isn't any rhyme or reason these work from a scripting standpoint.

PR #1140, opened 2024-07-16
Resolve most native joaat hashes

coalaura:
I collected any (gta-v related) strings I could find and built a joaat reverser with them. Turns out I also happened to find most native names this way (at least I believe they are correct since they seem to line up with the namespace and the documentation).

In case anyone is curious to use it: https://joaat.sh


PsychoShock:
While this is probably appreciate, I recommanded you __not__ do to this. Multiple reasons:
- Chance to break natives are 99%
- Impossible to review this correctly
- When you change/add new natives, try to do max 8 __per pr__

coalaura:
> You can still create multiple PR's for this big work!

@PsychoShock i definitely agree that this is hard to review correctly. I am not going to make multiple PRs as it would be way to much work to split it up tho and would probably take less time to review than to split it up. The joaat reverser i've built managed to reliably resolve the native joaat hashes tho, so maybe it can be of use for the project overall. Up to the maintainers to do with this PR as they please, just figured i'd share my findings considering the amount of names i (think) i found. :)

AvarianKnight:
Please refer to https://github.com/citizenfx/natives/pull/1154#issuecomment-2271979101

PR #1082, opened 2024-04-10
Update SetPedConfigFlag.md

alberttheprince:
Added the most up-to-date list of enum names to help use native documentation when setting flags.

Source:
https://alexguirre.github.io/rage-parser-dumps/dump.html?build=3095&game=gta5&search=configflags#ePedConfigFlags

Before you submit this PR, please make sure:

- You have read the contribution guidelines: Yes
- You include an example that validates your change: N/A
- Your English is grammatically correct: Yes

PR #1141, opened 2024-07-21
Update SetPedConfigFlag.md

FedericoDeBona:
Added all flags name (from https://docs.altv.mp/gta/articles/ped/flags.html)


PsychoShock (changes requested):
Hello,

While it's appreciate people who contribute (imo), copy/paste from another doc it's not a "good" standart. Couple problems with this:
- Name of the enum should be `ePedConfigFlags` (like it was before)
- It's missing __a lot__ of flags you have remove (even if name wasn't found, it's still relevant to have them)
- Really hard to review (if you make a mistake, the whole thing will need to be reverted)

For next time, you could only replace the names who haven't been found. Regarding this one, you can put back all the flags removed missing and set the correct enum name.

Thanks

AvarianKnight:
Closing in favor of https://github.com/citizenfx/natives/pull/1082

PR #1145, opened 2024-07-25
[PHYSICS - AddRope]: Updated Rope Type Info

JayPaulinCodes:
## Summary
Updated the AddRope native documentation to fix the incorrect information regarding the `ropeType` parameter and provided more details regarding the 8 possible rope types. Also included a simple example usage of the native in the form of a command.

## Validating Example(s)
- Video showcasing the 8 different rope types being spawned in: https://youtu.be/CkCVmEPQVgY
- Video showcasing the Lua native usage example: https://youtu.be/467hwsBaudg
- Code used for the validating examples:
```lua
local currentRope = nil
local propHandle = nil
local propHash = `prop_cs_fuel_nozle`
local pumpHash = `prop_gas_pump_1d`
local timeout = 1000
local pumpCoords = vector3(-2096.096436, -311.906891, 14.510918)
print("propHash", propHash)
print("pumpHash", pumpHash)

function spawnRopeOfType(ropeType)
    print("============================================")
    local pedHandle = PlayerPedId()
    local position = GetEntityCoords(pedHandle)
    print("ped", pedHandle, position)

    -- Create the prop
    if not HasModelLoaded(propHash) then
        local timer = 0
        while not HasModelLoaded(propHash) and timer < timeout do
            RequestModel(propHash)
            timer = timer + 1
            Citizen.Wait(1)
        end
    end
    propHandle = CreateObjectNoOffset(propHash, position.x, position.y, position.z, true, true, false)
    SetModelAsNoLongerNeeded(propHash)
    print("propHandle", propHandle)

    -- Attach the prop to ped
    local boneIndex = GetPedBoneIndex(pedHandle, 18905) -- SKEL_L_Hand
    print("boneIndex", boneIndex)
    AttachEntityToEntity(propHandle, pedHandle, boneIndex, 0.11, 0.02, 0.02, -15.0, -90.0, -80.0, false, false, false, false, 2, true)

    -- Create the rope
    local timer2 = 0
    while not RopeAreTexturesLoaded() and timer2 < timeout do
        RopeLoadTextures()
        timer2 = timer2 + 1
        Citizen.Wait(1)
    end

    currentRope = AddRope(
        0.0, -- x
        0.0, -- y
        0.0, -- z
        0.0, -- rotX
        0.0, -- rotY
        0.0, -- rotZ
        3.0, -- maxLength
        ropeType, -- ropeType
        1000.0, -- initLength
        0.0, -- minLength
        0.5, -- lengthChangeRate
        false, -- onlyPPU
        true, -- collisionOn
        true, -- lockFromFront
        1.0, -- timeMultiplier
        true, -- breakable
        0 -- unkPtr
    )
    print("currentRope", currentRope)
    ActivatePhysics(currentRope)

    -- Attach the rope to the prop
    local propPos = GetOffsetFromEntityInWorldCoords(propHandle, 0.0, -0.033, -0.195)
    print("propPos", propPos)
    local pumpHandle = GetClosestObjectOfType(pumpCoords.x, pumpCoords.y, pumpCoords.z, 10.0, pumpHash, false, true, true)
    print("pumpHandle", pumpHandle)
    local pumpCoords = GetEntityCoords(pumpHandle)
    print("pumpCoords", pumpCoords)
    
    AttachEntitiesToRope(currentRope, pumpHandle, propHandle, 
        pumpCoords.x, pumpCoords.y, pumpCoords.z,
        propPos.x, propPos.y, propPos.z,
        5.0, false, false, nil, nil)
end

RegisterCommand("ropetypes_dev_del", function(soruce, args)
    DeleteRope(currentRope)
    DeleteObject(propHandle)
end)

RegisterCommand("ropetypes_dev_spawn_0", function(soruce, args)
    print("ropetypes_dev_spawn_0")
    spawnRopeOfType(0)
end)

RegisterCommand("ropetypes_dev_spawn_1", function(soruce, args)
    print("ropetypes_dev_spawn_1")
    spawnRopeOfType(1)
end)

RegisterCommand("ropetypes_dev_spawn_2", function(soruce, args)
    print("ropetypes_dev_spawn_2")
    spawnRopeOfType(2)
end)

RegisterCommand("ropetypes_dev_spawn_3", function(soruce, args)
    print("ropetypes_dev_spawn_3")
    spawnRopeOfType(3)
end)

RegisterCommand("ropetypes_dev_spawn_4", function(soruce, args)
    print("ropetypes_dev_spawn_4")
    spawnRopeOfType(4)
end)

RegisterCommand("ropetypes_dev_spawn_5", function(soruce, args)
    print("ropetypes_dev_spawn_5")
    spawnRopeOfType(5)
end)

RegisterCommand("ropetypes_dev_spawn_6", function(soruce, args)
    print("ropetypes_dev_spawn_6")
    spawnRopeOfType(6)
end)

RegisterCommand("ropetypes_dev_spawn_7", function(soruce, args)
    print("ropetypes_dev_spawn_7")
    spawnRopeOfType(7)
end)


RegisterCommand("new_rope", function(soruce, args)

    -- Get the handle for the player's ped
    local pedHandle = PlayerPedId()

    -- Ensure that the rope textures are loaded
    local timer = 0
    while not RopeAreTexturesLoaded() and timer < 1000 do
        RopeLoadTextures()
        timer = timer + 1
        Citizen.Wait(0)
    end

    -- Get the coordinates for where the rope will be
    local ropePos = GetOffsetFromEntityInWorldCoords(pedHandle, 0.0, 2.0, 0.5)

    -- Create the rope
    local newRopeHandle = AddRope(
        ropePos.x, -- x
        ropePos.y, -- y
        ropePos.z, -- z
        0.0, -- rotX
        0.0, -- rotY
        0.0, -- rotZ
        10.0, -- maxLength
        1, -- ropeType
        10.0, -- initLength
        0.0, -- minLength
        1.0, -- lengthChangeRate
        false, -- onlyPPU
        false, -- collisionOn
        false, -- lockFromFront
        1.0, -- timeMultiplier
        false, -- breakable
        0 -- unkPtr
    )

end)
```

## Contribution Guidelines Acknowledgement
- [x] You have read the contribution guidelines
- [x] You include an example that validates your change
- [x] Your English is grammatically correct

### DIFF
--- a/PHYSICS/AddRope.md
+++ b/PHYSICS/AddRope.md
@@ -15,6 +15,56 @@ Rope does NOT interact with anything you attach it to, in some cases it make int
 Rope will sometimes contract and fall to the ground like you'd expect it to, but since it doesn't interact with the world the effect is just jaring.  
 ```
 
+## Rope Types
+There are 8 different rope types in the base game. Full rope data can be found in `ropedata.xml`.
+- Type: `rage__ropeThin`
+    - Rope Type Value: `0`
+    - Vertices: `1`
+    - Radius: `0.03`
+    - Texture Names: `rope` & `rope_n`
+
+- Type: `rage__ropeWire6`
+    - Rope Type Value: `1`
+    - Vertices: `4`
+    - Radius: `0.015`
+    - Texture Names: `steel_cable` & `steel_cable_n`
+
+- Type: `rage__ropeWire32`
+    - Rope Type Value: `2`
+    - Vertices: `32`
+    - Radius: `0.025`
+    - Texture Names: `steel_cable` & `steel_cable_n`
+
+- Type: `rage__ropeMesh`
+    - Rope Type Value: `3`
+    - Vertices: `6`
+    - Radius: `0.03`
+    - Texture Names: `rope` & `rope_n`
+
+- Type: `rage__ropeThinWire32`
+    - Rope Type Value: `4`
+    - Vertices: `32`
+    - Radius: `0.01`
+    - Texture Names: `rope` & `rope_n`
+
+- Type: `rage__ropeReins`
+    - Rope Type Value: `5`
+    - Vertices: `32`
+    - Radius: `0.005`
+    - Texture Names: `rope` & `rope_n`
+
+- Type: `rage__ropeThin4`
+    - Rope Type Value: `6`
+    - Vertices: `4`
+    - Radius: `0.03`
+    - Texture Names: `rope` & `rope_n`
+
+- Type: `rage__ropeWire64`
+    - Rope Type Value: `7`
+    - Vertices: `64`
+    - Radius: `0.025`
+    - Texture Names: `steel_cable` & `steel_cable_n`
+
 ## Parameters
 * **x**: Spawn coordinate X component.
 * **y**: Spawn coordinate Y component.
@@ -23,7 +73,7 @@ Rope will sometimes contract and fall to the ground like you'd expect it to, but
 * **rotY**: Rotation Y component.
 * **rotZ**: Rotation Z component.
 * **maxLength**: The maximum length the rope can droop.
-* **ropeType**: 1 to 4 are thick ropes. 5 and up are thin ropes. Ropes types defined in ropedata.xml. An invalid rope type such as 0 will crash the game.
+* **ropeType**: The zero-based index of the entry in the `ropedata.xml` file. *NOTE: An invalid rope type, such as `8`, will crash the game.*
 * **initLength**: The initial length of the rope.
 * **minLength**: The minimum length the rope can be.
 * **lengthChangeRate**: The speed in which the rope will wind if winding is started.
@@ -36,3 +86,45 @@ Rope will sometimes contract and fall to the ground like you'd expect it to, but
 
 ## Return value
 A script handle for the rope
+
+## Examples
+```lua
+RegisterCommand("new_rope", function(soruce, args)
+
+    -- Get the handle for the player's ped
+    local pedHandle = PlayerPedId()
+
+    -- Ensure that the rope textures are loaded
+    local timer = 0
+    while not RopeAreTexturesLoaded() and timer < 1000 do
+        RopeLoadTextures()
+        timer = timer + 1
+        Citizen.Wait(0)
+    end
+
+    -- Get the coordinates for where the rope will be
+    local ropePos = GetOffsetFromEntityInWorldCoords(pedHandle, 0.0, 2.0, 0.5)
+
+    -- Create the rope
+    local newRopeHandle = AddRope(
+        ropePos.x, -- x
+        ropePos.y, -- y
+        ropePos.z, -- z
+        0.0, -- rotX
+        0.0, -- rotY
+        0.0, -- rotZ
+        10.0, -- maxLength
+        1, -- ropeType
+        10.0, -- initLength
+        0.0, -- minLength
+        1.0, -- lengthChangeRate
+        false, -- onlyPPU
+        false, -- collisionOn
+        false, -- lockFromFront
+        1.0, -- timeMultiplier
+        false, -- breakable
+        0 -- unkPtr
+    )
+
+end)
+```
--- a/PHYSICS/AddRope.md
+++ b/PHYSICS/AddRope.md
@@ -17,53 +17,21 @@ Rope will sometimes contract and fall to the ground like you'd expect it to, but
 
 ### Rope Types
 There are 8 different rope types in the base game. Full rope data can be found in `ropedata.xml`.
-- Type: `rage__ropeThin`
-    - Rope Type Value: `0`
-    - Vertices: `1`
-    - Radius: `0.03`
-    - Texture Names: `rope` & `rope_n`
 
-- Type: `rage__ropeWire6`
-    - Rope Type Value: `1`
-    - Vertices: `4`
-    - Radius: `0.015`
-    - Texture Names: `steel_cable` & `steel_cable_n`
+Rope types:
 
-- Type: `rage__ropeWire32`
-    - Rope Type Value: `2`
-    - Vertices: `32`
-    - Radius: `0.025`
-    - Texture Names: `steel_cable` & `steel_cable_n`
-
-- Type: `rage__ropeMesh`
-    - Rope Type Value: `3`
-    - Vertices: `6`
-    - Radius: `0.03`
-    - Texture Names: `rope` & `rope_n`
-
-- Type: `rage__ropeThinWire32`
-    - Rope Type Value: `4`
-    - Vertices: `32`
-    - Radius: `0.01`
-    - Texture Names: `rope` & `rope_n`
-
-- Type: `rage__ropeReins`
-    - Rope Type Value: `5`
-    - Vertices: `32`
-    - Radius: `0.005`
-    - Texture Names: `rope` & `rope_n`
-
-- Type: `rage__ropeThin4`
-    - Rope Type Value: `6`
-    - Vertices: `4`
-    - Radius: `0.03`
-    - Texture Names: `rope` & `rope_n`
-
-- Type: `rage__ropeWire64`
-    - Rope Type Value: `7`
-    - Vertices: `64`
-    - Radius: `0.025`
-    - Texture Names: `steel_cable` & `steel_cable_n`
+```c
+enum ePhysicsRopeType {
+    RopeThin = 0, // Verticies: 1, Radius: 0.03, Textures: rope & rope_n
+    RopeWire6 = 1, // Verticies: 4, Radius: 0.015, Textures: steel_cable & steel_cable_n
+    RopeWire32 = 2, // Verticies: 32, Radius: 0.025, Textures: steel_cable & steel_cable_n
+    RopeMesh = 3, // Verticies: 6, Radius: 0.03, Textures: rope & rope_n
+    RopeThinWire32 = 4, // Verticies: 32, Radius: 0.01, Textures: rope & rope_n
+    RopeReins = 5, // Verticies: 32, Radius: 0.005, Textures: rope & rope_n
+    RopeThin4 = 6, // Verticies: 4, Radius: 0.03, Textures: rope & rope_n
+    RopeWire64 = 7 // Verticies: 64, Radius: 0.025, Textures: steel_cable & steel_cable_n
+}
+```
 
 ## Parameters
 * **x**: Spawn coordinate X component.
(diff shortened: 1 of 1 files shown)


PHYSICS/AddRope.md
```diff
@@ -35,4 +85,46 @@ Rope will sometimes contract and fall to the ground like you'd expect it to, but
...
+        10.0, -- initLength
+        0.0, -- minLength
+        1.0, -- lengthChangeRate
+        false, -- onlyPPU
+        false, -- collisionOn
+        false, -- lockFromFront
+        1.0, -- timeMultiplier
+        false, -- breakable
+        0 -- unkPtr
+    )
```
colistro123, line 127: Hmmm, why are we using a `timer` variable? This could cause the loop to break out if it ever goes over, which is unlikely, but still If this is going to take a while to load, it should be executed from a `thread` instead, i.e.

```lua
  -- Get the handle for the player's ped
  local pedHandle = PlayerPedId()
  -- Ensure that the rope textures are loaded
    
  while not RopeAreTexturesLoaded() do
    RopeLoadTextures()
  end
  
  -- Get the coordinates for where the rope will be
  local ropePos = GetOffsetFromEntityInWorldCoords(pedHandle, 0.0, 2.0, 0.5)
  -- Create the rope
  local newRopeHandle = AddRope(
      ropePos.x, -- x
      ropePos.y, -- y
      ropePos.z, -- z
      0.0, -- rotX
      0.0, -- rotY
      0.0, -- rotZ
      10.0, -- maxLength
      1, -- ropeType
      10.0, -- initLength
      0.0, -- minLength
      1.0, -- lengthChangeRate
      false, -- onlyPPU
      false, -- collisionOn
      false, -- lockFromFront
      1.0, -- timeMultiplier
      false, -- breakable
      0 -- unkPtr
  )
end)
```

That way it runs parallel to the main game thread.

colistro123 (changes requested):
Suggested some corrections and did some nitpicking, otherwise looks good to me for the most part.

PHYSICS/AddRope.md
```diff
@@ -35,4 +85,46 @@ Rope will sometimes contract and fall to the ground like you'd expect it to, but
 * **unkPtr**: Unknown pointer, always 0 in original scrips.
 
 ## Return value
-A script handle for the rope
\ No newline at end of file
+A script handle for the rope
+
+## Examples
+```lua
+RegisterCommand("new_rope", function(soruce, args)
```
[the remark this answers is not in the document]
JayPaulinCodes: Although it's true there is that Boolean, I've never actually ended up using that param in any of my scripts and have never had any issues so I don't see it being a requirement, however to align with how the method is represented in the rest of the documentation I would agree with you that it should be included.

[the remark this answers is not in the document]
colistro123: Yeah, it's mainly to align with how things are represented in the documentation 😊

PHYSICS/AddRope.md
JayPaulinCodes: Thank you for the input, I totally agree with what you have said here and will make an updated example to reflect this better. I've always used the timer variable as a way to protect against getting stuck in a loop forever, this I typically used more with things such as loading models or animations where there is the possibility of a bad input or a valid input which isn't able to be loaded for some reason, however probably isn't as applicable with rope textures. 

I will touch up this example with your feedback in mind shortly.

PHYSICS/AddRope.md
```diff
@@ -15,6 +15,56 @@ Rope does NOT interact with anything you attach it to, in some cases it make int
 Rope will sometimes contract and fall to the ground like you'd expect it to, but since it doesn't interact with the world the effect is just jaring.  
 ```
 
+## Rope Types
```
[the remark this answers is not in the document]
JayPaulinCodes: Wasn't aware of this, thank you for catching it!

PHYSICS/AddRope.md
```diff
@@ -23,7 +73,7 @@ Rope will sometimes contract and fall to the ground like you'd expect it to, but
 * **rotY**: Rotation Y component.
 * **rotZ**: Rotation Z component.
 * **maxLength**: The maximum length the rope can droop.
-* **ropeType**: 1 to 4 are thick ropes. 5 and up are thin ropes. Ropes types defined in ropedata.xml. An invalid rope type such as 0 will crash the game.
+* **ropeType**: The zero-based index of the entry in the `ropedata.xml` file. *NOTE: An invalid rope type, such as `8`, will crash the game.*
```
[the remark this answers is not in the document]
JayPaulinCodes: I would also agree there is definitely a better way to word this, I'll see what I can come up with that is more specific with the input you provided.

PHYSICS/AddRope.md
JayPaulinCodes: @colistro123 I have updated the example, it can be found [here](https://github.com/citizenfx/natives/pull/1145/files#diff-d378390d652ccf2c648ef059cf9b4b1b8fa20c619945d7bb502e047aba55d635R90-R137)

I moved the code into a thread as suggested and added a check after the while loop to ensure rope textures are loaded and check for that edge case. 

I opted to keep the timer and timeout for the while loop for the reason of preventing the very slim and odd edge case of the loop running forever, I don't see that ever happening however seeing as this is going in public documentation as an example I feel it's best to present code that doesn't have the potential of getting stuck in a infinite loop.

PHYSICS/AddRope.md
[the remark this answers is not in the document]
JayPaulinCodes: New wording can be found [here](https://github.com/citizenfx/natives/pull/1145/files#diff-d378390d652ccf2c648ef059cf9b4b1b8fa20c619945d7bb502e047aba55d635R76)

PHYSICS/AddRope.md
colistro123: I think that based on how the game internally loads those, that would be unlikely, but leave as is if you like. Point is that the game will always try to find an index to load those textures regardless.

image

Interestingly enough that function also gets called for specific vehicle fragment stuff, so it's unlikely for it to fail I suppose.

image

colistro123 (approved):
LGTM 😊

PHYSICS/AddRope.md
JayPaulinCodes: I see, that's some interesting info so thank you! I'll update the example accordingly

PHYSICS/AddRope.md
```diff
@@ -15,6 +15,56 @@ Rope does NOT interact with anything you attach it to, in some cases it make int
...
+    - Vertices: `4`
+    - Radius: `0.03`
+    - Texture Names: `rope` & `rope_n`
+
+- Type: `rage__ropeWire64`
+    - Rope Type Value: `7`
+    - Vertices: `64`
+    - Radius: `0.025`
+    - Texture Names: `steel_cable` & `steel_cable_n`
+
```
AvarianKnight, line 67: This seems like a lot of unneeded information, this could probably be an enum and still be fine

```c
enum ePhysicsRopeType {
    RopeInvalid = -1,
    RopeThin = 0,
    RopeDefault = 1,
    RopeDefault32 = 2,
    RopeDefaultWire = 3,
    RopeThinWire32 = 4,
    RopeReins = 5,
    RopeThin4 = 6,
}
```

PHYSICS/AddRope.md
```diff
@@ -35,4 +85,51 @@ Rope will sometimes contract and fall to the ground like you'd expect it to, but
...
+        -- Get the handle for the player's ped
+        local pedHandle = PlayerPedId()
+
+        -- Ensure that the rope textures are loaded
+        while not RopeAreTexturesLoaded() do
+            RopeLoadTextures()
+            Citizen.Wait(0)
+        end
+
+        if not RopeAreTexturesLoaded() then
```
AvarianKnight, line 104: This seems unneeded, there is no way to get here.

PHYSICS/AddRope.md
```diff
@@ -35,4 +85,51 @@ Rope will sometimes contract and fall to the ground like you'd expect it to, but
...
+
+    Citizen.CreateThread(function()
+        -- Get the handle for the player's ped
+        local pedHandle = PlayerPedId()
+
+        -- Ensure that the rope textures are loaded
+        while not RopeAreTexturesLoaded() do
+            RopeLoadTextures()
+            Citizen.Wait(0)
+        end
```
AvarianKnight, line 102:
```suggestion
        RopeLoadTextures()
        while not RopeAreTexturesLoaded() do
            Citizen.Wait(0)
        end
```

AvarianKnight (changes requested):

PHYSICS/AddRope.md
JayPaulinCodes: Part of the reason that I opted to display the extra info was because unless you go digging in the game files, it's not really found anywhere at all and knowing the amount of verts on a rope and the radius are, in my opinion, some useful info to have here as it will help developers find the rope that suits their needs the best for what they are working on.
With that said, if the standard is to omit this extra detail then I wouldn't be completely against changing to a enum, but I personally think it's important / useful enough info to keep here given the fact that it isn't documented elsewhere.

PHYSICS/AddRope.md
JayPaulinCodes: Yeah, totally right. I forgot about this part after I made the last change to the while loop, thanks for catching it!

PHYSICS/AddRope.md
AvarianKnight: I would generally say that it is unneeded and will clutter the page a lot, you could probably add the amount of verticies the rope uses to the end enum

PHYSICS/AddRope.md
```diff
@@ -35,4 +85,46 @@ Rope will sometimes contract and fall to the ground like you'd expect it to, but
 * **unkPtr**: Unknown pointer, always 0 in original scrips.
 
 ## Return value
-A script handle for the rope
\ No newline at end of file
+A script handle for the rope
+
+## Examples
+```lua
+RegisterCommand("new_rope", function(source, args, rawCommand)
```
AvarianKnight, line 92: Examples shouldn't be in a register command, please remove it and remove the extra indentation.

AvarianKnight (changes requested):

PHYSICS/AddRope.md
JayPaulinCodes: Alright then, thank you for the input and insight into the norms of the natives documentation, I will work on a update later today.

PHYSICS/AddRope.md
JayPaulinCodes: Out of curiosity, is there any reason why examples shouldn't be in commands? Just seems to me that an example command of how to use the native, in this case showing how to spawn a simple rope, would be a useful way to show an example of the native, so is there any particular reason or angle I may not be seeing as to why commands are not ideal examples?

AvarianKnight: Sorry for the late reply here, theres a few reasons here

1. The code is straight forward and only includes what it needs
2. The code becomes easier to copy & paste for people who are just starting and want to figure out what certain things do
3. We only have so much code space in the native docs currently, so the less space used by indention, the better.

PHYSICS/AddRope.md
```diff
@@ -23,7 +41,7 @@ Rope will sometimes contract and fall to the ground like you'd expect it to, but
 * **rotY**: Rotation Y component.
 * **rotZ**: Rotation Z component.
 * **maxLength**: The maximum length the rope can droop.
-* **ropeType**: 1 to 4 are thick ropes. 5 and up are thin ropes. Ropes types defined in ropedata.xml. An invalid rope type such as 0 will crash the game.
+* **ropeType**: The zero-based index of the entry in the `ropedata.xml` file. *NOTE: Using an index which does not exist will crash the game. As of right now, valid values are from `0` to `7` inclusive.*
```
AvarianKnight, line 44: `As of right now,` -> `On build x,`

PHYSICS/AddRope.md
```diff
@@ -15,6 +15,24 @@ Rope does NOT interact with anything you attach it to, in some cases it make int
 Rope will sometimes contract and fall to the ground like you'd expect it to, but since it doesn't interact with the world the effect is just jaring.  
 ```
 
+### Rope Types
```
AvarianKnight, line 18:
```suggestion

```
This is unneeded

PHYSICS/AddRope.md
```diff
@@ -15,6 +15,24 @@ Rope does NOT interact with anything you attach it to, in some cases it make int
 Rope will sometimes contract and fall to the ground like you'd expect it to, but since it doesn't interact with the world the effect is just jaring.  
 ```
 
+### Rope Types
+There are 8 different rope types in the base game. Full rope data can be found in `ropedata.xml`.
+
+Rope types:
+
```
AvarianKnight, line 22:
```suggestion

```
This is un-needed the enum is self documenting.

AvarianKnight (changes requested):

PHYSICS/AddRope.md
JayPaulinCodes: Good to know, thanks for the insight!

PR #1146, opened 2024-07-28
feat(docs/natives/RENDER_SCRIPT_CAMS): Renamed and explained the 4th …

koketoo:


CAM/RenderScriptCams.md
```diff
@@ -8,22 +8,16 @@ ns: CAM
...
-Setting ease to 1 will smooth the transition.  
-```
+Renders the cams the current script
+
+## Parameters
+* **render**: Whether to render the cams or not.
+* **ease **: If true it will make a smooth transition between the current camera and the new rendered camera.
+* **easeTime **: The in milliseconds for the transition to happen.
+* **easeCoordsAnim**: If false it will only do the smooth transition to the rotation and the coords will be set instantly.
+* **p4**: Doesn't seem to change anything.
```
PsychoShock, line 18: Param section should go after the `added param`.

PsychoShock, line 18: Try to use something more general and add what value is used by default in decompiled scripts.

PsychoShock (changes requested):

CAM/RenderScriptCams.md
koketoo: It's just that, if ease is true this param will change the behaviour in the transition

CAM/RenderScriptCams.md
```diff
@@ -8,22 +8,15 @@ ns: CAM
 void RENDER_SCRIPT_CAMS(BOOL render, BOOL ease, int easeTime, BOOL p3, BOOL p4);
```
colistro123, line 8: Where's `easeCoordsAnim`, which is `p3`? This is why the checks are failing.

colistro123 (changes requested):
Hey there, requested some changes.

CAM/RenderScriptCams.md
koketoo: Oh, now I see thanks

AvarianKnight (changes requested):

CAM/RenderScriptCams.md
```diff
@@ -5,25 +5,28 @@ ns: CAM
...
 
 ## Parameters
-* **render**: 
-* **ease**: 
-* **easeTime**: 
-* **p3**: 
-* **p4**: 
+* **render**: Whether to render the cams or not.
+* **ease **: If true it will make a smooth transition between the current camera and the new rendered camera.
+* **easeTime **: The in milliseconds for the transition to happen.
```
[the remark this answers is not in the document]
koketoo: Commited the suggestion, still not passing the build idk why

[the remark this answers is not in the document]
koketoo: Now yes, thanks!

PR #1153, opened 2024-08-06
[ENTITY] Resolve names

coalaura:
Resolved the names of a few natives in the ENTITY namespace. All done manually to ensure they are correct. Using https://joaat.sh/ to resolve the hashes.


ENTITY/GetEntityHeadingFromEulers.md
```diff
@@ -0,0 +1,17 @@
...
+// 0x846BF6291198A71E 0x9320E642
+float GET_ENTITY_HEADING_FROM_EULERS(Entity entity);
+```
+
+Gets the heading of the entity physics in degrees, which tends to be more accurate than just [`GET_ENTITY_HEADING`](#_0xE83D4F9BA2A38914). This can be clearly seen while, for example, ragdolling a ped/player.
+
+## Parameters
+* **entity**: The entity to get the heading from.
+
+## Return value
```
colistro123, line 17: Just so the return value is not empty, may I suggest:

```suggestion
## Return value
Returns a float representing the entity's heading from eulers.
```

colistro123 (commented):
LGTM

PR #1154, opened 2024-08-06
[GRAPHICS] Resolve names

coalaura:
Resolved the names of a few natives in the GRAPHICS namespace. All done manually to ensure they are correct. Using https://joaat.sh/ to resolve the hashes. Tried to keep the number of changes low for one PR.


AvarianKnight:
Thank you for the PR.

We're currently not accepting batch updates to natives. In the future these will be added though, so please be patient.

coalaura:
aight

PR #1156, opened 2024-08-08
tweak(natives/player): document that the recharge multiplier gets reset on player ped change

AvarianKnight:


colistro123 (commented):
LGTM, I haven't tested this, but it seems that calling [`NETWORK_RESURRECT_LOCAL_PLAYER`](https://docs.fivem.net/natives/?_0xEA23C49EAA83ACFB) might also reset this variable, especially if `bLeaveDeadPed` is passed as `true`.

I know it might be obvious, but respawning/dying could also cause this to reset (since it calls the `resurrectPlayerMethod` called by the native) when called from one of the game skeleton update functions such as `CGameLogic::Update`.

JayPaulinCodes:
Dittoing the comments by @colistro123 above but other than that 1 resurrect native that may be useful to add, all LGTM!

JayPaulinCodes (approved):
Dittoing the comments by @colistro123 above regarding other than that 1 resurrect native that may be useful to add, otherwise all LGTM!

AvarianKnight:
I can't replicate this behavior. 

```lua
CreateThread(function()
    SetEntityHealth(PlayerPedId(), 0)
    Wait(2000)
    local pos = GetEntityCoords(PlayerPedId())
    NetworkResurrectLocalPlayer(pos, 120.0, 0, true)
end)
```

then running `/crun SetPlayerHealthRechargeMultiplier(PlayerId(), 0.0)`, after being resurrected they still have the multiplier of 0.0

It should probably also be noted this *doesn't* seem to actually have a max value of `1.0` as it might imply from de-compiled scripts

Setting this to `SetPlayerHealthRechargeMultiplier(PlayerId(), 10.0)` works, and speeds up health regen by a lot, though I'm not sure why you would ever do this, it should probably be documented.

colistro123:
> I can't replicate this behavior.
> 
> ```lua
> CreateThread(function()
>     SetEntityHealth(PlayerPedId(), 0)
>     Wait(2000)
>     local pos = GetEntityCoords(PlayerPedId())
>     NetworkResurrectLocalPlayer(pos, 120.0, 0, true)
> end)
> ```
> 
> then running `/crun SetPlayerHealthRechargeMultiplier(PlayerId(), 0.0)`, after being resurrected they still have the multiplier of 0.0
> 
> It should probably also be noted this _doesn't_ seem to actually have a max value of `1.0` as it might imply from de-compiled scripts
> 
> Setting this to `SetPlayerHealthRechargeMultiplier(PlayerId(), 10.0)` works, and speeds up health regen by a lot, though I'm not sure why you would ever do this, it should probably be documented.

From looking at the native it seems that there doesn't seem to be any range limiting checks/clamping for how much you can set the multiplier to.

image

Couldn't replicate on my end either, I'm puzzled.

colistro123 (approved):

PR #1172, opened 2024-08-23
Update DisableAllControlActions.md

Mowayyy:
- Added enums instead of a hyperlink redirecting to the native ENABLE_ALL_CONTROL_ACTIONS.
- Added CAMERA_CONTROL enum which was unknown.

### DIFF
--- a/PAD/DisableAllControlActions.md
+++ b/PAD/DisableAllControlActions.md
@@ -5,9 +5,16 @@ ns: PAD
 
 ```c
 // 0x5F4B6931816E599B 0x16753CF4
-void DISABLE_ALL_CONTROL_ACTIONS(int padIndex);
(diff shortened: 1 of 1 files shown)


colistro123 (changes requested):
May I suggest leaving this file as-is?

Instead of modifying this file, why not simply update the `EnableAllControlActions.md` file? While this scenario is unlikely, it is preferable to centralize enums. This way, if updates are needed in the future, they can be made in a single file rather than in multiple locations.

Additionally, changing `padIndex` to `controlType` could impact the parameter naming conventions, [as most parameters for these kind of natives](https://github.com/search?q=repo%3Acitizenfx%2Fnatives%20padIndex&type=code) are currently named `padIndex`.

PR #1125, opened 2024-05-31
fix(images): use repo images

VIRUXE:
Images are now hosted with GitHub and pages use their relative URLs.

Not sure if this structure would be acceptable, but am open to suggestions.

PR #1056, opened 2024-03-11
Restore dead Imgur links

ItsJunction:
Many Imgur images were deleted at some point within the last year, this PR restores a fair few of them. Please let me know if I'm missing any not listed below.

Note this PR does NOT restore any dead links found in "unmodern" docs.

 The following links I could not find anything for:
- https://i.imgur.com/qLbXWcQ.png in AddBlipForArea.md
- https://i.imgur.com/0j7O7Rh.png in AddBlipForArea.md
- https://i.imgur.com/hS6ki7p.png in SetRadiusBlipEdge.md

PR #1174, opened 2024-08-23
tweak: change image host provider

userMacieG:


userMacieG:
https://github.com/citizenfx/fivem-docs/pull/475

colistro123:
This approach may not be ideal, as it introduces an additional repository as a dependency, specifically `fivem-docs`. Consequently, every time someone wants to add an image, they will also need to update that repository. This issue has been discussed [here](https://github.com/citizenfx/natives/pull/1056)

Just my two cents here, but a more efficient solution could involve using a serverless [AWS Lambda function](https://docs.aws.amazon.com/lambda/latest/dg/getting-started.html) (which can be implemented in Python, Node.js or any other supported language) in conjunction with an S3 bucket. This setup would allow users to upload images directly and serve the URLs on server response, making it easier to integrate them into the native documentation without the need for updates to the `fivem-docs` repository.

userMacieG:
But for now just to have working images, this in my opinion is best approach

VIRUXE:
I think my implementation is better. https://github.com/citizenfx/natives/pull/1125

PR #1176, opened 2024-08-24
Fixed some natives images and other things

NeenGame:


AvarianKnight:
While removing the `Citizen.` and updating natives to use proper native names is fine, we will not accept PR's that use external image hosting.

NeenGame:
> While removing the `Citizen.` and updating natives to use proper native names is fine, we will not accept PR's that use external image hosting.

Okay what should i use for the images? I don't know where to upload it, if you can tell me where to upload them, I will

PR #1183, opened 2024-08-29
feat(natives/ped): improve documentation for SET_PED_MIN_GROUND_TIME_FOR_STUNGUN

MadsLeander:
Improves the documentation for SET_PED_MIN_GROUND_TIME_FOR_STUNGUN by adding a more detailed description and examples.

The included example is a simple loop that sets the minimum ground time to 10 seconds whenever the player's ped changes.

### DIFF
--- a/PED/SetPedMinGroundTimeForStungun.md
+++ b/PED/SetPedMinGroundTimeForStungun.md
@@ -8,11 +8,84 @@ ns: PED
 void SET_PED_MIN_GROUND_TIME_FOR_STUNGUN(Ped ped, int ms);
 ```
 
-```
-Ped will stay on the ground after being stunned for at lest ms time. (in milliseconds)  
-```
+Overwrites the minimum time the ped will stay on the ground for after being stunned. Setting this while the ped is stunned will not alter the duration of the current stun (but will still effect future stunns).
+
+Unlike what this native's name suggests, it also effects the time for other "weapons", such as WEAPON_ELECTRIC_FENCE.
+
+Passing a negative value (usually -1) into the second parameter (ms) will reset the ground time back to default (ground time will then depend on the weapon that inflicts the stun).
 
 ## Parameters
-* **ped**: 
-* **ms**: 
+* **ped**: The ped to set the min ground time for
+* **ms**: The minimum time in milliseconds
 
+## Examples
+```lua
+-- This sets the minimum stun ground time for the player's ped to 10 seconds (and re-applies it if the player's ped changes)
+
+local currentPed = 0
+CreateThread(function()
+    while true do
+        local playerPed = PlayerPedId()
+
+        -- Checks if the player ped has changed
+        if currentPed ~= playerPed then
+            currentPed = playerPed
+
+            -- Sets the minimum stun ground time to 10 seconds
+            SetPedMinGroundTimeForStungun(currentPed, 10000)
+        end
+
+        Wait(1000)
+    end
+end)
+```
+
+```js
+// This sets the minimum stun ground time for the player's ped to 10 seconds (and re-applies it if the player's ped changes)
+
+let currentPed = 0;
+Delay = (ms) => new Promise(res => setTimeout(res, ms));
+
+setTick(async () => {
+    const playerPed = PlayerPedId();
+
+    // Checks if the player ped has changed
+    if (currentPed != playerPed) {
+        currentPed = playerPed;
+
+        // Sets the minimum stun ground time to 10 seconds
+        SetPedMinGroundTimeForStungun(currentPed, 10000);
+    };
+
+    await Delay(1000);
+});
+```
+
+```cs
+// This sets the minimum stun ground time for the player's ped to 10 seconds (and re-applies it if the player's ped changes)
+
+using static CitizenFX.Core.Native.API;
+// ...
+
+private int currentPed = 0;
+
+public Main() 
+{
+    Tick += OnTick;
+}
+
+private async Task OnTick()
+{
+    var playerPed = Game.PlayerPed.Handle;
+
+    // Checks if the player ped has changed
+    if (currentPed != playerPed)
+    {
+        currentPed = playerPed;
+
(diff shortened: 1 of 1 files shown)


PED/SetPedMinGroundTimeForStungun.md
```diff
@@ -8,11 +8,84 @@ ns: PED
...
+private int currentPed = 0;
+
+public Main() 
+{
+    Tick += OnTick;
+}
+
+private async Task OnTick()
+{
+    var playerPed = Game.PlayerPed.Handle;
```
JayPaulinCodes, line 79: Use the native instead of the wrapper, or add example for the native. See https://github.com/citizenfx/natives/pull/1148#discussion_r1702795868

PED/SetPedMinGroundTimeForStungun.md
```diff
@@ -8,11 +8,84 @@ ns: PED
 void SET_PED_MIN_GROUND_TIME_FOR_STUNGUN(Ped ped, int ms);
 ```
 
+Overwrites the minimum time the ped will stay on the ground for after being stunned. Setting this while the ped is stunned will not alter the duration of the current stun (but will still effect future stunns).
+
+Unlike what this native's name suggests, it also effects the time for other "weapons", such as WEAPON_ELECTRIC_FENCE.
```
JayPaulinCodes, line 13: Maybe let's change the native name then to better reflect the effect it has. Maybe something like SET_PED_MIN_GROUND_TIME_FOR_ STUN_EFFECT

PED/SetPedMinGroundTimeForStungun.md
```diff
@@ -8,11 +8,84 @@ ns: PED
...
+
+Passing a negative value (usually -1) into the second parameter (ms) will reset the ground time back to default (ground time will then depend on the weapon that inflicts the stun).
+
+## Parameters
+* **ped**: The ped to set the min ground time for
+* **ms**: The minimum time in milliseconds
```
JayPaulinCodes, line 19: Reword with "The minimum stun effect duration in milliseconds"

JayPaulinCodes (changes requested):

PED/SetPedMinGroundTimeForStungun.md
MadsLeander: I was unaware of that! Thanks for letting me know. I've pushed a commit to change it.

PED/SetPedMinGroundTimeForStungun.md
MadsLeander: I think that SET_PED_MIN_GROUND_TIME_FOR_STUNGUN is the actual native name (since it does not have an underscore prefix denoting that it is a community-named native). Maybe @AvarianKnight has some input here to confirm it or correct me on this?

AvarianKnight: We shouldn't change confirmed names.

JayPaulinCodes: Sounds good and makes sense :)

PED/SetPedMinGroundTimeForStungun.md
```diff
@@ -8,11 +8,84 @@ ns: PED
...
+    // Checks if the player ped has changed
+    if (currentPed != playerPed) {
+        currentPed = playerPed;
+
+        // Sets the minimum stun ground time to 10 seconds
+        SetPedMinGroundTimeForStungun(currentPed, 10000);
+    };
+
+    await Delay(1000);
+});
```
AvarianKnight, line 61:
```suggestion
```js
// This sets the minimum stun ground time for the player's ped to 10 seconds (and re-applies it if the player's ped changes)

let currentPed = 0;

setInterval(() => {
    const playerPed = PlayerPedId();

    // Checks if the player ped has changed
    if (currentPed !== playerPed) {
        currentPed = playerPed;

        // Sets the minimum stun ground time to 10 seconds
        SetPedMinGroundTimeForStungun(currentPed, 10000);
    };
}, 1000);
```

PED/SetPedMinGroundTimeForStungun.md
AvarianKnight, line 13:
```suggestion
**NOTE**: Unlike what the native name suggestios, this works on any weapon that has its `DamageType` in the `weapons.meta` set to `ELECTRIC`.
```
I need to verify this, just an guess currently

AvarianKnight (changes requested):

PED/SetPedMinGroundTimeForStungun.md
MadsLeander: An interval does indeed seem more appropriate here. Thanks for correcting me!

PED/SetPedMinGroundTimeForStungun.md
MadsLeander: I implemented Avarian's suggestion, but this nevertheless would have been an improvement over the original.

PED/SetPedMinGroundTimeForStungun.md
MadsLeander: I'll leave this unresolved until it is confirmed. I'll also take a look at it sometime this afternoon.

MadsLeander: Okay, after some testing, it seems correct that this native affects all damage from weapons where the `DamageType` is set to `ELECTRIC`. I have tried this with `WEAPON_ELECTRIC_FENCE`, `WEAPON_STUNROD`, `WEAPON_STUNGUN` and `WEAPON_STUNGUN_MP`. Custom (add-on) weapons also behaved the same when the `DamageType` was set to `ELECTRIC`. I could not find a way to test `VEHICLE_WEAPON_PLAYER_LASER` and `VEHICLE_WEAPON_ENEMY_LASER`, but I presume those are just leftovers from R*.

I am pretty confident that your suggested change is indeed correct, so I am going to commit it. If you discover something that contradicts these findings, please let me know.

PR #1207, opened 2024-09-28
[SHAKE_GAMEPLAY_CAM] Add all camera shakes

coalaura:
added missing camera shakes and formatted them in an easier to use enum. Let me know if it should be a plain list instead since its not theoretically an enum thats being used.


CAM/ShakeGameplayCam.md
```diff
@@ -8,22 +8,101 @@ ns: CAM
...
-LARGE_EXPLOSION_SHAKE  
-MEDIUM_EXPLOSION_SHAKE  
-SMALL_EXPLOSION_SHAKE  
-ROAD_VIBRATION_SHAKE  
-SKY_DIVING_SHAKE  
-VIBRATE_SHAKE  
+Shakes the gameplay cam with the specified shake name and intensity.
+
+```cpp
+enum eCameraShakes
```
JayPaulinCodes, line 14: Honestly, most other places which utilize enums do so with some form of number value attached to it as the parameter value where here it is a string parameter, in this current state using an enum could suggest that there is some tie to a numeric value when in reality there are just string inputs. 

Personally, I would leave this as a list of strings in a code block for styling.

JayPaulinCodes (changes requested):

CAM/ShakeGameplayCam.md
coalaura: sounds good, fixed

JayPaulinCodes (approved):
Looking better to me. I'll let other comment on the exact convention that we use for displaying these strings as I'm not 100% sure if the const like you have it is the convention. 

Other than that, all LGTM

PeekerJ:
When is this going to be merged?

PR #1209, opened 2024-10-02
Correct name and documentation for _PLACE_OBJECT_ON_GROUND_PROPERLY_2

coalaura:


OBJECT/PlaceObjectOnGroundOrObjectProperly.md
```diff
@@ -0,0 +1,17 @@
...
+aliases: ["0xD76EEEF746057FD6", "_PLACE_OBJECT_ON_GROUND_PROPERLY_2"]
+---
+## PLACE_OBJECT_ON_GROUND_OR_OBJECT_PROPERLY
+
+```c
+// 0xD76EEEF746057FD6
+BOOL PLACE_OBJECT_ON_GROUND_OR_OBJECT_PROPERLY(Object object);
+```
+
+Places an object on the ground or another object.
```
JayPaulinCodes, line 12: Seeing that we don't provide a coordinate set to place it at and assumingely it would use the object's X and Y position, maybe consider a wording such as: _Places an object on either the ground or on top of an object, whichever has a larger z coordinate at the object's x and y position._

JayPaulinCodes (changes requested):

OBJECT/PlaceObjectOnGroundOrObjectProperly.md
coalaura: I believe the game does a shapetest down. The difference between this one and `PLACE_OBJECT_ON_GROUND_PROPERLY` is this one sets the flag to include objects in the shapetest. I'll update the description accordingly

PR #1217, opened 2024-10-25
all speech names and params

coalaura:
Completed the list of speech names and params with all speech names and params i was able to find.


coalaura:
the list ended up growing larger than expected so i moved it to pastebin

AvarianKnight:
We shouldn't link to external sources, please make a PR to fivem-docs under game-references with this as a list.

PR #1218, opened 2024-10-27
update network natives

coalaura:
updated some control natives documentations


AvarianKnight:
These seem a little... weird. Not going to risk putting these into the native repository.

coalaura:
> These seem a little... weird. Not going to risk putting these into the native repository.

What do you mean with weird?

PR #1228, opened 2024-12-13
Update InterpolateCamWithParams.md

RocwoDev:
float camRotX,float camRotY -> float camRotX, float camRotY


AvarianKnight:
This will automatically be fixed whenever native updates happen

PR #1229, opened 2024-12-17
Update TaskGotoEntityOffset and TaskGotoEntityOffsetXY

coalaura:
Tested it aswell and it all seems to line up roughly


coalaura:
hold on there is more junk in that branch, lemme redo this

PR #1258, opened 2025-04-12
Update GetVehicleClass.md

killer305654:


AvarianKnight:
This doesn't seem accurate, trailers return as `11` -> `Utility` even on newer builds. If you found situations where it returns `23` please reopen this with examples.

Issue #1246, opened 2025-03-07
Fix zones list documentation


MichaelCoding25:
In the native [GetZoneFromNameId](https://github.com/citizenfx/natives/blob/edb875218ae1ce511c8c2aa564c655957a8a770f/ZONE/GetZoneFromNameId.md) the list of names that can be used is incorrect and currently always results in a value of `-1` being returned.

The correct list is in the popzone.ipl GTA file in the left-most column.

[The List](https://gist.github.com/MichaelCoding25/11455d704d030630156c3e742ee13e5e)

---

In the native [GetNameOfZone](https://github.com/citizenfx/natives/blob/edb875218ae1ce511c8c2aa564c655957a8a770f/ZONE/GetNameOfZone.md) the list of zones is not updated.

The updated list I got in the game files: `update\update.rpf\common\data\levels\gta5\popzone.ipl`

What needs to be changed in the more updated list from GetNameOfZone?
- Remove `SANAND` as it's not documented anywhere including game files
- Add `BAYTRE`: Baytree Canyon
- Add `GALLI`: Galileo Park
- Add `OBSERV`: Galileo Observatory

This is the updated list:

zone_name | zone_desc
-- | --
AIRP | Los Santos International Airport
ALAMO | Alamo Sea
ALTA | Alta
ARMYB | Fort Zancudo
BANHAMC | Banham Canyon Dr
BANNING | Banning
BAYTRE | Baytree Canyon
BEACH | Vespucci Beach
BHAMCA | Banham Canyon
BRADP | Braddock Pass
BRADT | Braddock Tunnel
BURTON | Burton
CALAFB | Calafia Bridge
CANNY | Raton Canyon
CCREAK | Cassidy Creek
CHAMH | Chamberlain Hills
CHIL | Vinewood Hills
CHU | Chumash
CMSW | Chiliad Mountain State Wilderness
CYPRE | Cypress Flats
DAVIS | Davis
DELBE | Del Perro Beach
DELPE | Del   Perro
DELSOL | La Puerta
DESRT | Grand Senora Desert
DOWNT | Downtown
DTVINE | Downtown   Vinewood
EAST_V | East Vinewood
EBURO | El Burro Heights
ELGORL | El Gordo Lighthouse
ELYSIAN | Elysian Island
GALFISH | Galilee
GALLI | Galileo Park
GOLF | GWC and Golfing Society
GRAPES | Grapeseed
GREATC | Great Chaparral
HARMO | Harmony
[...]

PR #1260, opened 2025-05-23
Update GetZoneFromNameId and GetNameOfZone

LokiLeiche:
These changes are based on Issue #1246. 

**GetZoneFromNameId**
In this native the docs page included a list of the zone abbreviations and their full names which is completely irrelevant for this native. It takes a zone string ID as listed in the firts column of `update\update.rpf\common\data\levels\gta5\popzone.ipl` as argument and returns the zone ID as integer, which is just the index of that zone in the popzone.ipl starting at 1.
The list being there is misleading as it causes people to think, the native converts the abreviations to the full are name which seemingly is also what the issue author thought but just isn't true.

**GetNameOfZone**
As the issue author stated, 3 zones that are listed in the popzone.ipl file are missing here, so I added them. Credits to @MichaelCoding25 for listing these. I verified these zones exist and work by making sure they are returned by the native when in the area configured by popzone.ipl. The issue author also said the zone `SANAND` should be removed since it's never used anywhere but I couldn't verify that so I left it for now.

Both of these have basically not been touched since the initial commit 7 years ago so they were probably never properly manually checked.


ZONE/GetNameOfZone.md
```diff
@@ -64,6 +66,7 @@ MTJOSE = Mount Josiah
 MURRI = Murrieta Heights  
 NCHU = North Chumash  
 NOOSE = N.O.O.S.E  
+NOOSE = Galileo Observatory
```
MichaelCoding25, line 69: Should be OBSERV and not NOOSE

LokiLeiche: Ah yeah, seems like I forgot to replace that after copy pasting... Sorry and thanks for pointing it out

MichaelCoding25:
Looks good and thanks for actually doing this.
One thing, maybe it's better in the ZONE/GetZoneFromNameId.md to have a list of these? because I feel most people won't wanna have to get software like openiv to get the actual file

LokiLeiche:
> Looks good and thanks for actually doing this. One thing, maybe it's better in the ZONE/GetZoneFromNameId.md to have a list of these? because I feel most people won't wanna have to get software like openiv to get the actual file

Thanks to you for doing the research for it. not sure about the list in GetZoneFromNameId. If you mean the old list that I removed, I feel like it's unnecessary since that is not the zone name used in the native so it's kinda misleading. If you mean a new zone list like the gist you had in your issue, that's way too long for adding it to the native documentation directly, but maybe link the gist or a pastebin with that list? idk

MichaelCoding25:
> > Looks good and thanks for actually doing this. One thing, maybe it's better in the ZONE/GetZoneFromNameId.md to have a list of these? because I feel most people won't wanna have to get software like openiv to get the actual file
> 
> Thanks to you for doing the research for it. not sure about the list in GetZoneFromNameId. If you mean the old list that I removed, I feel like it's unnecessary since that is not the zone name used in the native so it's kinda misleading. If you mean a new zone list like the gist you had in your issue, that's way too long for adding it to the native documentation directly, but maybe link the gist or a pastebin with that list? idk

Yeah Maybe using the gist or making a pastebin would be good. I just updated the gist to include the ids and descriptions of the zones

LokiLeiche:
Thanks @technetium-cfx for the review, I opened a PR in fivem-docs where I added the list as game reference and updated the link to it here, therefore only merge this with my new PR. https://github.com/citizenfx/fivem-docs/pull/554#issue-3335366242

If there's anything else, let me know.

PR #1270, opened 2025-07-22
Update ToggleUsePickupsForPlayer.md

IlMelons:
Improved description of the native with a practical example of use in lua

### DIFF
--- a/OBJECT/ToggleUsePickupsForPlayer.md
+++ b/OBJECT/ToggleUsePickupsForPlayer.md
@@ -9,6 +9,13 @@ aliases: ["0x616093EC6B139DD9"]
 void _TOGGLE_USE_PICKUPS_FOR_PLAYER(Player player, Hash pickupHash, BOOL toggle);
 ```
 
+Toggles the possibility for the player to drop the weapon pickups after death
+
+```lua
+local pickup = GetHashKey("PICKUP_WEAPON_ADVANCEDRIFLE")
+ToggleUsePickupsForPlayer(PlayerId(), pickup, false)
+```
+
 ```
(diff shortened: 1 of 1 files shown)


OBJECT/ToggleUsePickupsForPlayer.md
```diff
@@ -9,6 +9,13 @@ aliases: ["0x616093EC6B139DD9"]
 void _TOGGLE_USE_PICKUPS_FOR_PLAYER(Player player, Hash pickupHash, BOOL toggle);
 ```
 
+Toggles the possibility for the player to drop the weapon pickups after death
```
AvarianKnight, line 12: This should go at the very bottom, and have an `## Examples` header before it so it will show properly in the docs.

OBJECT/ToggleUsePickupsForPlayer.md
```diff
@@ -21,7 +28,6 @@ SET_PLAYER_*
 ```
```
AvarianKnight, line 28: You can remove this old information

AvarianKnight (changes requested):

OBJECT/ToggleUsePickupsForPlayer.md
IlMelons: Done, I didn't knew how markdown was managed, I searched other natives as example to try to format it correctly

OBJECT/ToggleUsePickupsForPlayer.md
IlMelons: Done, I keeped only the first lines of the example as you marked, is that right or you meant all the codeblock?

PR #1286, opened 2025-09-19
Update CreateMoneyPickups.md to include optional method parameter

MartinKleinschrot:
Added optional method parameters


AvarianKnight:
That is sadly, not how this works. There's no native tooling to handle this.

PR #1295, opened 2025-11-11
Clarify EXTEND_WORLD_BOUNDARY_FOR_PLAYER usage and call frequency

M981:
Not sure if this is useful to specify, but 'ExtendWorldBoundaryForPlayer' doesn’t need to be called in a loop and can be used to prevent vehicle engines from dying out of bounds.


AvarianKnight:
Any new comments shouldn't be in the code blocks, they're meant for native docs imports only.

PR #1297, opened 2025-11-29
Update SetPedEnveffScale.md

McSwampy:
Better explination on the function


AvarianKnight:
You shouldn't enclose comments in codeblocks (```) as these are only meant to be used by imports of native docs

PR #1323, opened 2026-02-26
Fix description for SET_TRANSITION_TIMECYCLE_MODIFIER

ekaliclua:
## What was changed
Clarified the `transition` parameter description for SET_TRANSITION_TIMECYCLE_MODIFIER.

Replaced vague wording ("Appears to be half-seconds (?)") with a more accurate value (1/3 of seconds).

## Validation
Tested in-game by applying different transition values and measuring the duration.
The timing matches ~0.33 seconds per unit.

No parameter or return types were modified.

### DIFF
--- a/GRAPHICS/SetTransitionTimecycleModifier.md
+++ b/GRAPHICS/SetTransitionTimecycleModifier.md
@@ -12,5 +12,5 @@ This native doesn't work like [`SetWeatherTypeTransition`](#_0x578C752848ECFA0C)
 
 ## Parameters
 * **modifierName**: The name of a timecycle
-* **transition**: The speed to transition to `modifierName`. Appears to be half-seconds (?)
+* **transition**: The speed to transition to `modifierName`. Appears to be 1/3 of seconds.
 


technetium-cfx:
Would be curious to know what led to the 1/3 seconds per unit. Looking into this, it seems to be related to SET_TIMECYCLE_MODIFIER_STRENGTH instead.

colistro123:
> Would be curious to know what led to the 1/3 seconds per unit. Looking into this, it seems to be related to SET_TIMECYCLE_MODIFIER_STRENGTH instead.

The second parameter is the time taken, known as `time`. `1.0/3` is the end result of the `fTransitionSpeed` once it settles into the game engine via this native call setter, where `fTransitionSpeed` would be an internal game variable off the `CTimeCycle` vmt. The modifier strength can be set via a native named `SetTimeCycleModifierStrength`, although I cannot seem to reference the global variable it sets on v323 as used in any sections of the game, perhaps due to it being obscured by compiler optimization level.

image

Now with that out of the way, what we for sure know is that `1.0` is the default `fModifierStrength` (although I ambiguously named it `fTransitionStrength` in my documented native method) setup via `CTimeCycle::Init` and `3` is the second parameter known as `time`.

image


These are somewhat easy to cross-reference due to the unique nature of the offsets.

image

PR #1337, opened 2026-06-13
Clarify IS_PED_GETTING_INTO_A_VEHICLE return behavior

phongvants123:
Clarified return values for IsPedGettingIntoAVehicle function.


phongvants123:
Tested in-game.

IS_PED_GETTING_INTO_A_VEHICLE returns false while the ped is still walking toward the vehicle and before the door is opened.

It returns true after the vehicle door has been opened.

Screenshots attached below demonstrate both states.
image
image

